PR #79, opened 2015-01-01
removed custom agent packages variable

adamlavie:


suites/Vagrantfile
```diff
@@ -50,7 +50,6 @@ def build_env(test_config)
     end
```
ran-z: overall looks ok, but you've missed some code that needs to be removed at system_tests.sh, check it out.

PR #80, opened 2015-01-04
CFY-1801 added changes to the nodecellar test to make it more generic.

yaelpeled:


cosmo_tester/test_suites/test_blueprints/nodecellar_test.py
```diff
@@ -166,6 +170,18 @@ def post_uninstall_assertions(self):
         except ConnectionError:
             pass
 
+    def get_expected_nodes_count(self):
```
dankilman: Consider using properties instead of getter methods.
For example, instead of 
``` python
def get_host_expected_runtime_properties(self):
    return ['ip', 'networks']
```
which will be used like this:
``` python
self.get_host_expected_runtime_properties()
```

You can have something like this
``` python
@property
def host_expected_runtime_properties(self):
    return ['ip', 'networks']
```
which can then be used like a simple attribute access (with no method invocation syntax)
``` python
self.host_expected_runtime_properties
```

PR #89, opened 2015-01-26
Heal test

simon-bar:
Autoheal test adjusted to the most recent code


cosmo_tester/test_suites/test_blueprints/nodecellar_test_autoheal.py
```diff
@@ -139,14 +130,20 @@ def post_autoheal_assertions(self, before_state, after_state):
...
                             {'ExampleCollector': {}})
+            # TODO: Remove this link replacement when suitable
+            imports = patch._get_object_by_path('imports')
+            imports = [
+                imp if 'types.yaml' not in imp
+                else 'https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager/master'
+                     '/resources/rest-service/cloudify/types/types.yaml'
+                for imp in imports
+            ]
+            patch.set_value('imports', imports)
```
dankilman: There is no need to patch the imports. the current types.yaml already contains your recent changes

cosmo_tester/test_suites/test_blueprints/nodecellar_test_autoheal.py
```diff
@@ -139,14 +130,20 @@ def post_autoheal_assertions(self, before_state, after_state):
 
     def modify_blueprint(self):
         with YamlPatcher(self.blueprint_yaml) as patch:
-            patch.merge_obj('workflows', self.AUTOHEAL_WORKFLOW_YAML)
             patch.merge_obj('groups', self.AUTOHEAL_GROUP_YAML)
-            # FIXME: this is a workaround for multiple collectors starting
-            # multiple auto-heal workflows
-            patch.set_value("node_types.nodecellar\\.nodes\\.MonitoredServer."
+            patch.merge_obj("node_types.nodecellar\\.nodes\\.MonitoredServer."
```
dankilman: The node cellar example already has many collectors configured.
it would be best if instead of adding the example collector, we would simply update properties.service (in the groups section) to be something like "cpu\\.avg.*" (or whatever will be a valid regex that will catch exactly one metric)

cosmo_tester/test_suites/test_blueprints/nodecellar_test_autoheal.py
simon-bar: Yes, I have just compared it with http://www.getcloudify.org/spec/cloudify/3.2m3/types.yaml and it is identical. I will remove this TODO, it doesn't pass flake8 anyway

cosmo_tester/test_suites/test_blueprints/nodecellar_test_autoheal.py
simon-bar: That's a good idea

PR #123, opened 2015-03-30
added security nodecellar test with bootstrap

noakup:


cosmo_tester/test_suites/test_security/secured_nodecellar_test.py
```diff
@@ -0,0 +1,113 @@
...
+                    'name': 'password',
+                    'implementation': 'flask_securest.'
+                                      'authentication_providers.password:'
+                                      'PasswordAuthenticator',
+                    'properties': {
+                        'password_hash': 'plaintext'
+                    }
+                },
+                {
+                    'name': 'token',
```
dankilman: If this test does not actually test the token provider, it should not be added here

cosmo_tester/test_suites/test_security/secured_nodecellar_test.py
```diff
@@ -0,0 +1,113 @@
...
+    def _set_credentials_env_vars(self):
+        os.environ[CLOUDIFY_USERNAME_ENV] = TEST_CFY_USERNAME
+        os.environ[CLOUDIFY_PASSWORD_ENV] = TEST_CFY_PASSWORD
+
+    def _running_env_setup(self):
+        cfy = CfyHelper(cfy_workdir=self.workdir)
+        self.env.management_ip = cfy.get_management_ip()
+        self.client = CloudifyClient(
+            self.env.management_ip,
+            user=os.environ[CLOUDIFY_USERNAME_ENV],
```
dankilman: If the username/password constants are already available is see no need in reading the from the evironment as they were set only one function before. The constants can be used directly

cosmo_tester/test_suites/test_security/secured_nodecellar_test.py
```diff
@@ -0,0 +1,113 @@
+########
+# Copyright (c) 2014 GigaSpaces Technologies Ltd. All rights reserved
```
dankilman: 2015

cosmo_tester/test_suites/test_security/secured_nodecellar_test.py
```diff
@@ -0,0 +1,113 @@
...
+                           task_retries=5,
+                           install_plugins=False)
+        self.addCleanup(self.cfy.teardown)
+
+    def _set_credentials_env_vars(self):
+        os.environ[CLOUDIFY_USERNAME_ENV] = TEST_CFY_USERNAME
+        os.environ[CLOUDIFY_PASSWORD_ENV] = TEST_CFY_PASSWORD
+
+    def _running_env_setup(self):
+        cfy = CfyHelper(cfy_workdir=self.workdir)
```
dankilman: self.cfy is already configured (from base TestCase), no need to create a new cfy client.

suites/suites/suites.yaml
```diff
@@ -336,28 +336,6 @@ handler_configurations:
       <<: *lab_openstack_inputs
       keystone_tenant_name: system-tests/system-test-4
 
-  lab_openstack_system_tests4_secured:
```
dankilman: There should also be a diff that removes/modifies the suites that uses `lab_openstack_system_tests4_secured` handler configuration, no?
I mean, I suppose you want modify the existing suite definition + add a new test group with the security test

cosmo_tester/test_suites/test_security/secured_nodecellar_test.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def _update_manager_blueprint(self, prop_path, new_value):
+        with util.YamlPatcher(self.test_manager_blueprint_path) as patch:
+            patch.set_value(prop_path, new_value)
+
+    def _bootstrap(self):
+        self.cfy.bootstrap(blueprint_path=self.test_manager_blueprint_path,
+                           inputs_file=self.test_inputs_path,
+                           task_retries=5,
+                           install_plugins=False)
```
dankilman: Let's discuss `install_plugins` tomorrow, I also have a bug in the rest_plugins test where it is also `False`.
Generally, when running tests locally, you do want this to be `False` so that it doesn't override packages installed in your environment, but when run as part of the automation, some plugins may not be installed (such as the fabric plugin) so the install_plugins is required there.

cosmo_tester/test_suites/test_security/secured_nodecellar_test.py
```diff
@@ -0,0 +1,113 @@
...
+
+import os
+from path import path
+
+from cosmo_tester.framework import util
+from cosmo_tester.framework.cfy_helper import CfyHelper
+from cosmo_tester.test_suites.test_blueprints.nodecellar_test \
+    import OpenStackNodeCellarTestBase
+
+from cloudify_rest_client.client import CloudifyClient
```
dankilman: the `cloudify_rest_client` and `cloudify_cli` imports should be the second part (between the standard lib imports the the system tests imports), because even though they are cloudify packages, they are 3rd party with regards to the cloudify-system-tests project

PR #142, opened 2015-04-16
added the handler configuration to the suites.yaml and manager inputs

EarthmanT:


suites/configurations/inputs-aws-ec2.yaml
```diff
@@ -0,0 +1,18 @@
+aws_access_key_id: ''
+aws_secret_access_key: ''
+image_id: ''
+instance_type: ''
+manager_keypair_name: 'shared-systemt-tests-key'
+agent_keypair_name: 'shared-systemt-tests-key'
+
+use_existing_manager_group: false
+use_existing_agent_group: false
+use_existing_manager_keypair: false
```
dankilman: `use_existing_manager_keypair` and `use_existing_agent_keypair` should be `true`

suites/suites/suites.yaml
```diff
@@ -381,7 +389,18 @@ handler_configurations:
...
+  aws_ec2:
+    inputs: inputs-aws-ec2.yaml
+    handler: ec2_handler
+    external: *aws_ec2_plugin_external
+    manager_blueprint: aws-ec2/aws-ec2-manager-blueprint.yaml
+    inputs_override:
+      aws_access_key_id: '{{system_tests_aws_access_key_id}}'
+      aws_secret_access_key: '{{system_tests_aws_secret_access_key}}'
+
+  ################################################################################
```
dankilman: there  should be spaces at the start if this separator `###...` instead of ` ####...`

suites/suites/suites.yaml
```diff
@@ -381,7 +389,18 @@ handler_configurations:
...
+
+  aws_ec2:
+    inputs: inputs-aws-ec2.yaml
+    handler: ec2_handler
+    external: *aws_ec2_plugin_external
+    manager_blueprint: aws-ec2/aws-ec2-manager-blueprint.yaml
+    inputs_override:
+      aws_access_key_id: '{{system_tests_aws_access_key_id}}'
+      aws_secret_access_key: '{{system_tests_aws_secret_access_key}}'
```
dankilman: aren't you missing the `image_id` and `instance_type` inputs?
In order to avoid duplicating these here and in the properties, look at the `variables` section on top, and add these with reasonable names there, then in the properties and inputs sections you can use `'{{my_var_name'}}` instead of hard coded `ami-3cf8b154` (for example)

suites/configurations/inputs-aws-ec2.yaml
EarthmanT: check

suites/suites/suites.yaml
EarthmanT: ok

suites/suites/suites.yaml
EarthmanT: ok I did both things

dankilman:
So all that remains is creating the user in aws, uploading the public key to it with the correct name and adding the access key and secret key of this user to quickbuild

EarthmanT:
Discussed with Barak. He’s thinks adding another account is too much maintenance. He wants all of us to talk today.

PR #147, opened 2015-04-20
CFY-2209 added some utility methods to be used by tests

iliapolo:


cosmo_tester/framework/util.py
```diff
@@ -124,6 +128,29 @@ def get_actual_keypath(env, keypath, raise_on_missing=True):
     return keypath
 
 
+def render_template_to_file(template_path, file_path=None, **values):
```
dankilman: This function could be simplified to
```
rendered = render_template(template_path, **values)
return content_to_file(renderered, file_path)
```

cosmo_tester/framework/util.py
```diff
@@ -253,3 +280,10 @@ def _get_parent_obj_prop_name_by_path(self, prop_path):
     @staticmethod
     def _raise_illegal(prop_path):
         raise RuntimeError('illegal path: {0}'.format(prop_path))
+
+
+def generate_password():
+    length = 13
+    chars = string.ascii_letters + string.digits + '!@#$%^&*()'
+    random.seed = (os.urandom(1024))
```
dankilman: not sure what the added value for changing the seed here is?
It won't be any more random than it already is

cosmo_tester/framework/util.py
```diff
@@ -20,11 +20,15 @@
...
+import string
+import random
+import tempfile
 
 import jinja2
-from path import path
 import yaml
+from path import path
 from itsdangerous import base64_encode
+from jinja2 import Template
```
dankilman: I think it'd be better to import `jinja2` and do `jinja2.Template` where necessary

PR #153, opened 2015-04-22
TFY-15 Added manager override

EarthmanT:
I am trying to open the port 8086 in aws ec2 manager during system tests.

### DIFF
--- a/suites/suites/suites.yaml
+++ b/suites/suites/suites.yaml
@@ -113,6 +113,13 @@ templates:
       port: 8086
       cidr: 0.0.0.0/0
 
+  - &aws_ec2_manager_blueprint_sg_override
+    'node_templates.management_security_group.properties.rules[append]':
+      ip_protocol: tcp
+      from_port: 8086
+      to_port: 8086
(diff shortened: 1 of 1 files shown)


suites/suites/suites.yaml
```diff
@@ -121,6 +121,17 @@ templates:
     <<: *manager_blueprint_commercial_override
     node_templates.manager.properties.cloudify.workflows.task_retries: 100
 
+  - &aws_ec2_manager_blueprint_override
+    <<: *manager_blueprint_override
```
dankilman: `manager_blueprint_override` is good, which is something that we actually forgot initially.

suites/suites/suites.yaml
```diff
@@ -121,6 +121,17 @@ templates:
...
+    <<: *aws_ec2_manager_blueprint_sg_override
+
+  - &aws_ec2_manager_blueprint_sg_override
```
dankilman: `&aws_ec2_manager_blueprint_sg_override` should be placed before `aws_ec2_manager_blueprint_override`.
Specifically, put it next to the other security group override templates

PR #158, opened 2015-04-29
CFY-2644 added mongo collector assertions

iliapolo:


cosmo_tester/test_suites/test_blueprints/nodecellar_test.py
```diff
@@ -194,7 +220,7 @@ def repo_url(self):
 
     @property
     def repo_branch(self):
-        return None  # will use git_helper.clone default branch
+        return 'temp-mongo-broken'
```
iliapolo: revert

PR #167, opened 2015-05-07
CFY-2640-Add-monitoring-test-to-system-test-for-Hello-World-Example

gilzellner:
CFY-2640-Add-monitoring-test-to-system-test-for-Hello-World-Example

### DIFF
--- a/cosmo_tester/framework/assertions.py
+++ b/cosmo_tester/framework/assertions.py
@@ -1,0 +1,14 @@
+########
+# Copyright (c) 2015 GigaSpaces Technologies Ltd. All rights reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+#    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    * See the License for the specific language governing permissions and
+#    * limitations under the License.
--- a/cosmo_tester/framework/assertions.py
+++ b/cosmo_tester/framework/assertions.py
@@ -12,3 +12,26 @@
 #    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #    * See the License for the specific language governing permissions and
 #    * limitations under the License.
+
+from influxdb import InfluxDBClient
+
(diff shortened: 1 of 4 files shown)


cosmo_tester/framework/assertions.py
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+#    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    * See the License for the specific language governing permissions and
+#    * limitations under the License.
+
+
+from influxdb import InfluxDBClient
+
+
+def assert_general_deployment_data(self, influx_client):
```
ran-z: is this method supposed to be called from outside this module as well? if not, it should have an underscore prefix

note that if this method is a public method (i.e. no underscore prefix), its name is rather ambiguous; the name made (some) sense before, when it was a private method, since it was called from a "assert_monitoring_.." method. If you check the nodecellar_test code, you can see that  the assert_monitoring_data_exists calls both this and another method which verifies mongodb collector data - which is the reason for the original, still a-bit ambiguous name.

cosmo_tester/framework/assertions.py
```diff
@@ -0,0 +1,39 @@
...
+        influx_client.query('select * from /^{0}\./i '
+                            'where time > now() - 5s'
+                            .format(self.test_id))
+    except NameError as e:
+        self.fail('monitoring events list for deployment with ID {0} were'
+                  ' not found on influxDB. error is: {1}'
+                  .format(self.deployment_id, e))
+
+
+def assert_monitoring_data_exists(self):
```
ran-z: you've extracted the method form the hello-world test, but havent done so from the nodecellar test. you'll want to be extracting the code from there as well - both cause otherwise the refactoring doesn't serve much purpose :), but also because it might make your refactoring change  - and indeed i believe that in this case you'll want to modify the structure of your methods somewhat.

gilzellner: that is for a separate jira issue and branch, no ?

cosmo_tester/framework/assertions.py
gilzellner: fixed

cosmo_tester/framework/assertions.py
ran-z: nope
the refactoring is part of this issue - and the refactoring includes all its parts, including extracting from other code.
i guess if you want to be a purist, you could create a different jira dedicated specifically for refactoring and do the entire there there (including the first half of the refactoring, which you've already done) - but in my opinion it's just cumbersome and not really all that useful

gilzellner: done

PR #171, opened 2015-05-17
CFY-2641-Add-outputs-test-to-system-test-for-Hello-World-Example

gilzellner:


cosmo_tester/test_suites/test_blueprints/hello_world_bash_test.py
```diff
@@ -93,6 +93,10 @@ def _run(self, image_name, user, is_existing_deployment=False):
         floating_ip_id, neutron, nova, sg_id, server_id =\
             self._verify_deployment_installed()
         self.assert_monitoring_data_exists()
+        floating_ip_dict = neutron.show_floatingip(floating_ip_id)
+        expected_output = '{{u\'http_endpoint\': u\'http://{0}:8080\'}}' \
```
ran-z: 8080 --> use the property from the server node instead (already done in this very module, you can look how its done there)

cosmo_tester/framework/testenv.py
```diff
@@ -265,6 +265,13 @@ def setUpClass(cls):
     def tearDownClass(cls):
         pass
 
+    def wait_for_expected_outputs(self, expected_outputs, timeout=3):
```
ran-z: why is the "wait" required?

cosmo_tester/framework/testenv.py
```diff
@@ -265,6 +265,13 @@ def setUpClass(cls):
...
+        def assertion():
+            outputs = self.client.deployments.outputs.get(self.test_id)
+            outputs = outputs['outputs']
+            self.assertEqual(str(expected_outputs), str(outputs))
```
ran-z: why the use of str()? if these objects are dictionary there are different methods for comparing those

cosmo_tester/test_suites/test_blueprints/hello_world_bash_test.py
gilzellner: ok

cosmo_tester/framework/testenv.py
gilzellner: the other places I saw getting outputs had delays. but those were monitoring. I guess it can be removed if you think it is a bad idea.

cosmo_tester/framework/testenv.py
gilzellner: ok, will do.

cosmo_tester/framework/testenv.py
ran-z: monitoring data is not the same as output; monitoring data is sent continuously by the diamond agents, whereas a deployment's outputs (the "cloudify meaning of outputs" in this case :)) should be available right after running the install workflow.
i think the wait/timeouts should be removed.

cosmo_tester/framework/testenv.py
ran-z: (FYI, if they are indeed dictionary, string comparing might be more than bad practice and can theoretically fail, as the order of elements in dictionary is not guaranteed)

gilzellner: it is not really a dictionary in the sense there is only one value, but you are correct since some values may be added in the future.

cosmo_tester/framework/testenv.py
gilzellner: םל

cosmo_tester/framework/testenv.py
```diff
@@ -265,6 +265,11 @@ def setUpClass(cls):
     def tearDownClass(cls):
         pass
 
+    def assertOutputs(self, expected_outputs):
```
ran-z: no camelcase ^_^

cosmo_tester/framework/test_cases.py
```diff
@@ -20,8 +20,9 @@
 
 class MonitoringTestCase(TestCase):
 
-    def _assert_general_deployment_data(self, influx_client):
-
+    def assert_deployment_monitoring_data_exists(self, deployment_id):
```
iliapolo: the deployment_id should have a default value of self.test_id

cosmo_tester/framework/testenv.py
```diff
@@ -265,6 +265,11 @@ def setUpClass(cls):
     def tearDownClass(cls):
         pass
 
+    def assert_outputs(self, expected_outputs):
+        outputs = self.client.deployments.outputs.get(self.test_id)
```
iliapolo: here again, dont assume self.test_id is the deployment id, allow for passing it and assign self.test_id as the default value of the deployment id

cosmo_tester/test_suites/test_blueprints/hello_world_bash_test.py
```diff
@@ -92,7 +92,14 @@ def _run(self, image_name, user, is_existing_deployment=False):
 
         floating_ip_id, neutron, nova, sg_id, server_id =\
             self._verify_deployment_installed()
-        self.assert_monitoring_data_exists()
+        self.assert_deployment_monitoring_data_exists(self.test_id)
```
iliapolo: then here you dont have to pass anything to the assert_bla_bla...because this test does not use a deployment id of its own

cosmo_tester/test_suites/test_blueprints/nodecellar_test.py
```diff
@@ -171,7 +171,7 @@ def _assert_monitoring_data_exists(self):
         client = InfluxDBClient(self.env.management_ip, 8086, 'root', 'root',
                                 'cloudify')
         self._assert_mongodb_collector_data(client)
-        self.assert_monitoring_data_exists()
+        self.assert_deployment_monitoring_data_exists(self.deployment_id)
```
iliapolo: and here, the self.deployment_id property doesn't really exist. the test doesn't specify a deployment id when deploying the blueprint, meaning that self.test_id will be used.

PR #176, opened 2015-06-10
Max retries softlayer

yaelpeled:


suites/suites/suites.yaml
```diff
@@ -126,7 +126,6 @@ templates:
 
   - &softlayer_manager_blueprint_override
     <<: *manager_blueprint_commercial_override
-    node_templates.manager.properties.cloudify.workflows.task_retries: 35
```
dankilman: because this manager blueprint override is now identical to the `manager_blueprint_commercial_override` manager blueprint override, we can remove this entirely and reference `manager_blueprint_commercial_override` in the softlayer handler configuration directly

PR #225, opened 2015-09-09
CFY-3337 fixed some tests

nirbir:


suites/configurations/inputs-datacentred-openstack.yaml
```diff
@@ -12,7 +12,7 @@ external_network_name: external
 use_existing_manager_keypair: true
 use_existing_agent_keypair: true
 manager_server_name: systemt-cloudify-manager
-manager_server_user: ubuntu
+ssh_user: ubuntu
```
ran-z: also do this for the other openstack inputs files:
mirantis
hp
lab

PR #229, opened 2015-09-22
CFY-3629 Added support to installing plugins on rest service.

max-orlov:


max-orlov:
Fixed the test for install-plugins to work with the new manager blueprint.

PR #234, opened 2015-10-01
CFY-3337 removed irrelevant tests

nirbir:


suites/suites/suites.yaml
```diff
@@ -675,11 +675,6 @@ tests:
     tests:
       - cosmo_tester/test_suites/test_blueprints -e puppet -e chef -e docker -e windows
 
-  openstack_blueprints_persistence_recovery:
```
dankilman: do not actually delete these 'test' entries but rather make sure that are not referenced from the`tests_suites` section

suites/suites/suites.yaml
```diff
@@ -675,11 +675,6 @@ tests:
...
-    tests:
-      - cosmo_tester/test_suites/test_blueprints/docker_persistence_test.py
```
dankilman: make the commit message + PR message state that this change is only relevant for m5 and should not be merged to master (m6) as well

PR #235, opened 2015-10-01
created shell script for quickbuild to run to generate summary report

noakup:


suites/tests_report_requirements.txt
```diff
@@ -0,0 +1,6 @@
+sh
```
dankilman: these requirements don't seem right. you are not making use of `sh`, `path.py`, `nose` and `xmltodict`.
you are making use of `lxml`, `gspread` and `oauth2client` which don't seem to appear in the requirements. 
i suspect the 2 last modules probably belong to some single google python API library or something

PR #247, opened 2015-10-14
Test scaling

simon-bar:
0 instances -> 1 instance -> 0 instances scale test


idanmo:
An integration test was added.

PR #273, opened 2015-11-06
Add protocol for security group rule

kostya13:


suites/suites/suites.yaml
```diff
@@ -142,6 +142,7 @@ templates:
           destination: internal
           destination_port: 8086
           action: allow
+          protocol: TCP
```
dankilman: isn't this the default?

kostya13: For vcloud default value is "ANY".

PR #278, opened 2015-11-10
CFY 4064 integrate vcloud nodecellar test to sys tests env

ran-z:


suites/suites/suites.yaml
```diff
@@ -658,6 +666,23 @@ handler_configurations:
...
+    handler: vcloud_handler
+    env: vcloud
+    external: *vcloud_plugin_external
+    manager_blueprint: new/vcloud-manager-blueprint.yaml
+    properties: vcloud_handler_props
+    manager_blueprint_override: *vcloud_manager_blueprint_override
+    inputs_override:
+      vcloud_username: '{{system_test_vcloud_username}}'
+      vcloud_password: '{{system_test_vcloud_password}}'
+      vcloud_service_type: '{{system_test_vcloud_service_type}}'
```
dankilman: shouldn't `vcloud_service_type` and the two that follow it be defined in the `variables` section? are they private in any way to justify putting them in quickbuild?
(also, add the ones that are in quickbuild to the upper section where these are documented with a default `_`

PR #281, opened 2015-11-10
CFY-3360 added a nodecellar 3.2.1 backwards test

idanmo:


cosmo_tester/test_suites/backwards/test_nodecellar_backwards_compatibility/nodecellar_321_backwards_test.py
```diff
@@ -0,0 +1,44 @@
...
+            'agent_user': 'ubuntu'
+        }
+
+    @property
+    def repo_branch(self):
+        return 'tags/3.2.1'
+
+    @property
+    def entrypoint_node_name(self):
+        return 'nodecellar_ip'
```
max-orlov: It appears that the correct node name is nodecellar_floatingip (from https://github.com/cloudify-cosmo/cloudify-nodecellar-example/blob/3.2.1-build/openstack-blueprint.yaml)

PR #301, opened 2015-11-20
CFY-4065:  timeout customization for nodecellar

0lvin:


suites/suites/suites.yaml
```diff
@@ -379,6 +379,8 @@ handler_properties:
 
   vsphere_handler_props:
     template: systemtests-ubuntu-1404
+    agent_user: giga
+    timeout: 3600
```
ran-z: think it should be named "execution_timeout" instead

0lvin: changed

PR #303, opened 2015-11-24
Added some AWS EC2 related constants used by rhel cli test (in cloudi…

simon-bar:
…fy-packager)

### DIFF
--- a/suites/suites/suites.yaml
+++ b/suites/suites/suites.yaml
@@ -41,8 +41,10 @@ variables:
   datacentred_openstack_centos_7_image_id: 74ff4015-aee1-4e02-aaa8-1c77b2650394
   datacentred_openstack_windows_image_name: Windows Server 2012 R2 Standard
 
+  rhel_7_image_user: ec2-user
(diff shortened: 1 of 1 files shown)


suites/suites/suites.yaml
```diff
@@ -389,8 +391,10 @@ handler_properties:
   aws_ec2:
     ubuntu_trusty_image_id: '{{aws_ec2_eu_central_1_ami_ubuntu_trusty_image_id}}'
     centos_7_image_id: '{{aws_ec2_eu_central_1_ami_centos_7_image_id}}'
+    aws_ec2_eu_west_1_ami_rhel_7_image_id: '{{aws_ec2_eu_west_1_ami_rhel_7_image_id}}'
```
idanmo: Please name it rhel_7_image_id.

Once we'll run the tests on different regions we will have a different set of properties per region and the image_id property name should be exactly the same so that we won't be limited to running a certain test on a certain region.

simon-bar: Yep, this is actually what I was writing about on flowdoc ;) The name is good for a variable, but we should have one clean name here and rely on specific configs

PR #304, opened 2015-11-25
CFY-3967 implementation: two managers based snapshots test

pawelcodilime:


cosmo_tester/test_suites/test_snapshots/snapshots_multiple_managers_test.py
```diff
@@ -0,0 +1,135 @@
+import random
```
idanmo: Please add a license header.

cosmo_tester/test_suites/test_snapshots/snapshots_multiple_managers_test.py
```diff
@@ -0,0 +1,135 @@
...
+
+from cloudify_rest_client.exceptions import CloudifyClientError
+
+from cosmo_tester.framework.cfy_helper import CfyHelper
+from cosmo_tester.framework.util import create_rest_client
+from cosmo_tester.test_suites.test_blueprints.nodecellar_test import (
+    OpenStackNodeCellarTest)
+
+
+class TwoManagersTest(OpenStackNodeCellarTest):
```
idanmo: Aren't there supposed to be tests in this class?

cosmo_tester/test_suites/test_snapshots/snapshots_multiple_managers_test.py
```diff
@@ -0,0 +1,135 @@
...
+    It is required that there is one additional manager defined in the
+    handler configuration.
+    """
+
+    def setUp(self):
+        super(TwoManagersTest, self).setUp()
+        configuration = self.env.handler_configuration
+        self.workdir2 = tempfile.mkdtemp(prefix='cloudify-testenv-')
+
+        if 'additional_managers_ips' in self.env.handler_configuration:
```
idanmo: I think that it would be best if the test would bootstrap both managers not using a manager bootstrapped by the test environment as this package will be used for all snapshots tests and running the tests on the same manager which is used for restoring snapshots is probably a bad idea.

cosmo_tester/test_suites/test_snapshots/snapshots_multiple_managers_test.py
pawelcodilime: There is one test in the class OpenStackNodeCellarTest which is sufficient for this purpose. The class TwoManagersTest only overrides some methods.

cosmo_tester/test_suites/test_snapshots/snapshots_multiple_managers_test.py
pawelcodilime: So basically I shouldn't touch the manager which is available through self.cfy and self.client and just bootstrap two another managers?

cosmo_tester/test_suites/test_snapshots/__init__.py
```diff
@@ -14,16 +14,7 @@
...
-requests.packages.urllib3.disable_warnings()
-
-
-def setUp():
-    bootstrap()
-
-
-def tearDown():
-    teardown()
+requests.packages.urllib3.disable_warnings(
```
idanmo: Perhaps put this code in the testenv module so it will be relevant for all tests?

simon-bar: This is something that I proposed Sandy in our e-mail discussion about silencing those warnings, maybe let him decide?

idanmo: The discussion is relevant for the existing solution we applied in CLI. I suggest to try your fix in system tests as we have a lot of clutter in test logs because of this warning.

simon-bar: Ok, no problem. You want to put it here:

 cloudify-system-tests/cosmo_tester/framework/testenv.py

right? Or maybe even here  cloudify-system-tests/cosmo_tester/framework/__init__.py ?

simon-bar: We will add it to testenv.py

cosmo_tester/test_suites/test_snapshots/snapshots_multiple_managers_test.py
```diff
@@ -0,0 +1,176 @@
...
+    the second manager, checks whether nodecellar is not running indeed and
+    tears down those managers.
+    """
+
+    def setUp(self):
+        super(TwoManagersTest, self).setUp()
+        configuration = self.env.handler_configuration
+        self.workdir2 = tempfile.mkdtemp(prefix='cloudify-testenv-')
+
+        if 'second_manager_ip' in self.env.handler_configuration:
```
idanmo: Hi,
Seems like this code is not relevant... [second_manager_ip in handler_configuration]

PR #310, opened 2015-11-30
Join two plugins tests into one.

pawelcodilime:


cosmo_tester/test_suites/test_managed_plugins/managed_plugins_test.py
```diff
@@ -94,8 +84,11 @@ def _verify_plugin_can_be_used_in_blueprint(self):
             self.cfy.delete_blueprint(self.test_id)
             shutil.rmtree(blueprint_path)
 
-    def _delete_all_plugins(self):
-        for plugin in self.client.plugins.list():
+    def _delete_remote_plugin_if_exists(self):
+        plugins = self.client.plugins.list(package_name=TEST_PACKAGE_NAME)
+        if plugins:
+            # one result is expected
```
simon-bar: If one result is expected, we should assert it I think..
Maybe we should do it by plugin's ID?

geokala:

PR #319, opened 2015-12-07
Single host snapshot test is based on Hello World now instead of Node…

simon-bar:
…Cellar


simon-bar:
Don't merge this PR yet please ;)

PR #336, opened 2015-12-27
CFY-3717 active directory test

adamlavie:


cosmo_tester/resources/blueprints/active-directory/active-directory-blueprint.yaml
```diff
@@ -0,0 +1,103 @@
+tosca_definitions_version: cloudify_dsl_1_2
+
+description: >
+  This blueprint defines a EC2 VM created using Cloudify's AWS plugin
```
noakup: a EC2 => an EC2

cosmo_tester/resources/blueprints/active-directory/active-directory-blueprint.yaml
```diff
@@ -0,0 +1,103 @@
...
+      agent_config:
+        install_method: none
+      image_id: { get_input: image_id }
+      instance_type: { get_input: instance_type }
+      parameters: 
+        user_data: |
+          <powershell>
+            $Computer = [ADSI]"WinNT://$Env:COMPUTERNAME,Computer"
+            $bob = $Computer.Create("User", "bob")
+            $bob.SetPassword("b0b_Aa123456!")
```
noakup: why are you adding alice, bob and clair to the administrators group? only alice should be there, while bob is a manager and clair is a viewer. that way we can differentiate between the users. if you only need an administrator you can remove bob and clair.

cosmo_tester/resources/blueprints/active-directory/scripts/sleep.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+echo $1
```
noakup: hmmpphhh... what am I missing? how is this sleep?..

cosmo_tester/resources/blueprints/active-directory/scripts/wait_for_active_directory_port.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+HOST_IP=$public_ip
+PORT=$port
+TIMEOUT_MIN="$(($timeout_minutes * 60))"
```
noakup: why multiply by 60 if both values are in minutes?

cosmo_tester/resources/blueprints/active-directory/scripts/wait_for_active_directory_port.sh
```diff
@@ -0,0 +1,14 @@
...
+function wait_for_port
+{
+    /bin/netcat -z -w$3 $1 $2
+    if [ $? -ne 0 ] && $fail_on_timeout; then
```
noakup: 1. what if fail_on_timeout is set to False and the netcat command fails not because of the timeout?
it should probably be something like:
if return_code is not 0:
  if fail_on_timeout:
    echo and exit 1
  else:
    echo and do something else
2. the error message writes "minutes" but it's actually in seconds

cosmo_tester/resources/blueprints/active-directory/active-directory-blueprint.yaml
adamlavie: 1. These are active directory users. The group they belong to is meaningless for the test since this test only tests authentication against the active directory and not authorization. Adding them to the Administrators group is just a quick way for having them added as users to the active directory later on.
2. It is supported by amazon and probably also by the plugin. I'll need to consult with Trummel

cosmo_tester/resources/blueprints/active-directory/scripts/sleep.sh
adamlavie: this was definitely not meant to be pushed

cosmo_tester/resources/blueprints/active-directory/scripts/wait_for_active_directory_port.sh
adamlavie: netcat accepts input in seconds. Since I want to convert input from the user, I multiply his input (stated in the bp as timeout_minutes) by 60 to get the result in minutes

adamlavie: netcat accepts input in seconds. Since I want to convert input from the user, I multiply his input (stated in the bp as timeout_minutes) by 60 to get the result in minutes

cosmo_tester/resources/blueprints/active-directory/scripts/wait_for_active_directory_port.sh
adamlavie: 1. The netcat command does not raise an error and there for will not terminate the script with exit code 1.
2. The error message is fine the way it is.

cosmo_tester/test_suites/test_security/ad_auth_test.py
```diff
@@ -0,0 +1,152 @@
...
+                                    ' --no-index --find-links=wheels/ --pre'
+                }
+            }
+
+    # Use file based userstore
+    def get_userstore_driver(self):
+        return {
+            'implementation': 'flask_securest.userstores.file_userstore:'
+                              'FileUserstore',
+            'properties': {
```
adamlavie: userstore_file_path

cosmo_tester/test_suites/test_security/auth_test_base.py
```diff
@@ -0,0 +1,550 @@
...
+        # start authorization assertions
+        self._assert_blueprint_operations()
+        self._assert_deployment_operations()
+        self._assert_execution_operations()
+
+    def _test_authentication_without_authorization(self, assert_token=None):
+        self._assert_valid_credentials_authenticate()
+        self._assert_invalid_credentials_fails()
+        self._assert_empty_credentials_fails()
+        self._assert_no_credentials_or_token_fails()
```
adamlavie: verify

cosmo_tester/test_suites/test_security/authentication_test.py
```diff
@@ -12,96 +12,32 @@
...
-        settings = {
-            SECURITY_PROP_PATH + '.enabled': self.get_enabled(),
-        }
-
-        authentication_providers = self.get_authentication_providers()
-        if authentication_providers:
-            settings[SECURITY_PROP_PATH + '.authentication_providers'] = \
-                authentication_providers
-
-        settings[SECURITY_PROP_PATH + '.authorization_provider'] = ''
```
adamlavie: missed this one!

cosmo_tester/test_suites/test_security/security_test_base.py
```diff
@@ -35,6 +37,7 @@ def setup_secured_manager(self):
         self._copy_manager_blueprint()
         if self.get_ssl_enabled():
             self._handle_ssl_files()
+        self._update_userstore_file()
```
adamlavie: don't update whenm not using the file_userstore driver

cosmo_tester/test_suites/test_security/ad_auth_test.py
```diff
@@ -0,0 +1,152 @@
...
+    # The ldap authentication plugin requires openldap-devel.
+    def get_userdata(self):
+        return '#!/bin/bash sudo yum install openldap-devel -y'
+
+    # this test does not require the passwords be stored in the userstore
+    def get_userstore_users(self):
+        return [
+            {
+                'username': auth_test_base.ADMIN_USERNAME,
+                'groups': [
```
noakup: these groups don't exist in your active directory, so they should not be part of the userstore

cosmo_tester/resources/blueprints/active-directory/active-directory-blueprint.yaml
```diff
@@ -0,0 +1,103 @@
...
+            $clair = $Computer.Create("User", "clair")
+            $clair.SetPassword("cl@ir_Aa123456!")
+            $clair.SetInfo()
+            $alice = $Computer.Create("User", "alice")
+            $alice.SetPassword("@lice_Aa123456!")
+            $alice.SetInfo()
+            $group = [ADSI]"WinNT://./Administrators,group"
+            $group.Add("WinNT://bob,user")
+            $group.Add("WinNT://clair,user")
+            $group.Add("WinNT://alice,user")
```
noakup: what about dave? we need dave to verify a user without roles can't perform operations

cosmo_tester/test_suites/test_security/ad_auth_test.py
```diff
@@ -0,0 +1,152 @@
...
+                'username': auth_test_base.DEPLOYER_USERNAME,
+                'groups': [
+                    'cfy_deployers'
+                ]
+            },
+            {
+                'username': auth_test_base.VIEWER_USERNAME,
+                'groups': [
+                    'cfy_viewer'
+                ]
```
noakup: dave is missing?

cosmo_tester/test_suites/test_security/auth_test_base.py
```diff
@@ -0,0 +1,550 @@
+########
```
noakup: There is an essential problem with the way you merged the files. AuthenticationWithoutAuthorization was meant to verify the server's behavior when bootstrapping with an empty authorization section, while the authorization tests verify the server's behavior when an authorization section is part of the blueprint.
The way you merged them loses the first functionality (not using an authorization section at all).
What you want to do here is leave the two test files separately, allow the different bootstrap configurations to be tested, and just extract the assertions. Naming-wise, you should probably name the extracted assertions as "assert_authentication" and "assert_authorization". Furthermore, while the authentication tests use a rest client, authorization tests use cfy. the login method now does both, even though they are never both needed - either the test needs to init a rest client, or it need to set env vars. never both.

cosmo_tester/test_suites/test_security/authorization_tests.py
```diff
@@ -13,435 +13,25 @@
 #    * See the License for the specific language governing permissions and
 #    * limitations under the License.
```
noakup: undo - this file should be restored

cosmo_tester/test_suites/test_security/authentication_test.py
```diff
@@ -12,96 +12,32 @@
 #    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
noakup: undo - this file should be restored

cosmo_tester/test_suites/test_security/auth_test_base.py
```diff
@@ -0,0 +1,559 @@
...
+ADMIN_USERNAME = 'alice'
+ADMIN_PASSWORD = 'alice_password'
+DEPLOYER_USERNAME = 'bob'
+DEPLOYER_PASSWORD = 'bob_password'
+VIEWER_USERNAME = 'clair'
+VIEWER_PASSWORD = 'clair_password'
+NO_ROLE_USERNAME = 'dave'
+NO_ROLE_PASSWORD = 'dave_password'
+
+RUNNING_EXECUTIONS_MESSAGE = 'There are running executions for this deployment'
```
noakup: I think this constant isn't used anymore

cosmo_tester/test_suites/test_security/no_userstore_tests.py
```diff
@@ -40,7 +39,7 @@ def get_manager_blueprint_additional_props_override(self):
         shutil.copytree(src_plugin_dir,
                         self.test_manager_blueprint_path.dirname() /
                         CUSTOM_AUTH_PROVIDER_PLUGIN)
-        return {PLUGINS_PROP_PATH: self.get_plugins_settings()}
+        return {REST_PLUGIN_PATH: self.get_plugins_settings()}
```
noakup: are you sure this works? cause now you're adding 'user_custom_auth_provider' under `node_templates.rest_service.properties` and not under `node_templates.rest_service.properties.plugins`

cosmo_tester/test_suites/test_security/security_test_base.py
```diff
@@ -118,8 +134,39 @@ def _update_manager_blueprint(self):
...
+        with util.YamlPatcher(self.test_userstore_file_path) as patch:
+            for key, value in userstore_settings.items():
+                patch.set_value(key, value)
+
     def get_manager_blueprint_additional_props_override(self):
-        return {}
+        overrides = {}
+        rest_plugins = self.get_rest_plugins()
+        if rest_plugins:
+            overrides = {'{0}.plugins'.format(REST_PLUGIN_PATH): rest_plugins}
```
noakup: As mentioned in the previous comment, `REST_PLUGIN_PATH` should contain `.plugins` in it (meaning its value should remain as `PLUGINS_PROP_PATH` was, maybe just renamed if your prefer).
Also, if you only use `MANAGER_HOST_PARAMETERS` for userdata, change it to include `.user_data` and rename it to `USER_DATA_PATH`. That way you can avoid the explicit prefixing added in this method.

cosmo_tester/test_suites/test_security_ec2/ad_auth_test.py
```diff
@@ -0,0 +1,161 @@
+########
+# Copyright (c) 2015 GigaSpaces Technologies Ltd. All rights reserved
```
noakup: I don't see a reason to put this test in a different module, since it is a security test and uses the same flow. maybe add it to the existing folder of tests?

cosmo_tester/test_suites/test_security_ec2/__init__.py
```diff
@@ -0,0 +1,26 @@
+########
```
noakup: given the next comment (see below), this file is redundant.

PR #338, opened 2015-12-31
Update of system test which performs migration from 3.2.1 to 3.3 usin…

xwombat:
…g snapshots.


cosmo_tester/test_suites/test_snapshots/snapshots_migration_from_321_to_33_test.py
```diff
@@ -385,9 +412,33 @@ def _install_hello_world_on_3_2_1(self):
...
+        )
+
+        hello_blueprint_path = os.path.join(hello_repo_path, 'blueprint.yaml')
+
+        with YamlPatcher(hello_blueprint_path) as patch:
+            patch.merge_obj(
+                'node_templates.security_group.interfaces',
+                {'cloudify.interfaces.lifecycle': {
+                    'create': {
+                        'implementation':
```
dankilman: no need to explicitly pass the implementation, `inputs` is enough

PR #343, opened 2016-01-10
setting env vars by overriding super

noakup:


cosmo_tester/test_suites/test_security/security_test_base.py
```diff
@@ -179,6 +179,10 @@ def _set_credentials_env_vars(self):
         os.environ[constants.CLOUDIFY_USERNAME_ENV] = TEST_CFY_USERNAME
         os.environ[constants.CLOUDIFY_PASSWORD_ENV] = TEST_CFY_PASSWORD
 
+    def _unset_credentials_env_vars(self):
+        del os.environ[constants.CLOUDIFY_USERNAME_ENV]
```
dankilman: it is preferable you do `os.environ.pop(constants.CLOUDIFY_USERNAME_ENV, None)` to protect against the case the env var was never updated

cosmo_tester/test_suites/test_security/security_test_base.py
```diff
@@ -189,9 +193,13 @@ def _running_env_setup(self):
         self.env.management_ip = self.cfy.get_management_ip()
         self.set_rest_client()
 
-        def clean_mgmt_ip():
+        def clear_mgmt_and_security_settings():
             self.env.management_ip = None
-        self.addCleanup(clean_mgmt_ip)
+            self.env.rest_client = None
+            self.cfy = None
```
dankilman: `cfy` and `client` are overridden by the base class `setUp` for every test so it is not really required cleaning them

PR #374, opened 2016-02-24
CFY-4746 Added info about network to which vm has to connect in test_…

xwombat:
…existing_vm test

### DIFF
--- a/cosmo_tester/test_suites/test_blueprints/test_existing_vm.py
+++ b/cosmo_tester/test_suites/test_blueprints/test_existing_vm.py
@@ -41,7 +41,7 @@ class ExistingVMTest(TestCase):
         agents_security_group = self.env.agents_security_group
         management_network_name = self.env.management_network_name
 
-        nova_client, _, _ = self.env.handler.openstack_clients()
+        nova_client, neutron_client, _ = self.env.handler.openstack_clients()
 
         self.logger.info('Creating keypair...')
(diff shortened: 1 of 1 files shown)


cosmo_tester/test_suites/test_blueprints/test_existing_vm.py
```diff
@@ -99,19 +104,28 @@ def create_keypair_and_copy_to_manager(self,
         fabric.api.put(local_path=key_file,
                        remote_path=remote_key_path)
 
+    def get_nics_by_network_name(self, network_name, neutron_client):
+        nets = neutron_client.list_networks()
```
ran-z: FYI, you could pass keyword arguments into this call and have the server filter out the networks it returns, so instead of looking for the one with the correct name on the client side, you could get back only the networks with the given name.

cosmo_tester/test_suites/test_blueprints/test_existing_vm.py
```diff
@@ -99,19 +104,28 @@ def create_keypair_and_copy_to_manager(self,
...
+
+        for net in nets['networks']:
+            if net['name'] == network_name:
+                return [{'net-id': net['id']}]
+
```
ran-z: I'd probably have the "raise" line here, rather than outside of the function, I just find it clearer that way, but this obviously doesn't matter much in this case.

PR #375, opened 2016-02-25
no longer assuming self.env.management_key_path and self.env.manageme…

nir0s:
…nt_user are declared


cosmo_tester/framework/testenv.py
```diff
@@ -321,10 +321,18 @@ def setUp(self):
         self.logger.setLevel(logging.INFO)
         self.logger.info('Starting test setUp')
         self.workdir = tempfile.mkdtemp(prefix='cosmo-test-')
+        if hasattr(self.env, 'management_user_name'):
```
dankilman: could be cleaned up by
``` python
management_user = getattr(self.env, 'management_user_name', None)
```

PR #377, opened 2016-03-06
windows tests name change

max-orlov:


suites/suites/suites.yaml
```diff
@@ -860,8 +860,8 @@ test_suites:
     tests:
       - psutil_plugin
 
-  openstack_windows_cli_package:
-    requires: [openstack, datacentred]
+  ec2_windows_cli_package:
+    requires: aws_ec2
```
max-orlov: Add to jenkins default before merge

PR #393, opened 2016-04-11
CFY-4607 cleanup: each test removes keypairs it creates

tehasdf:
Operations that can partially fail, and that need some cleanup
(eg. the install/uninstall workflows) need to have their cleanup
scheduled (using .addCleanup) BEFORE running them.

In case the operation (eg. install) fails partway, and throws an
exception, the .addCleanup call might be skipped because of the
exception, and the uninstall won't run, leaving the system with
a partially-executed workflow that was never cleaned up after.

This change makes sure that all .addCleanup calls are always scheduled
before the risky operation they're cleaning up after.

This also uses the new openstack plugin API for removing keypairs.


tehasdf:
this requires the new API proposed in cloudify-cosmo/cloudify-openstack-plugin#110

max-orlov:
There are some system test in [cloudify-package](https://github.com/cloudify-cosmo/cloudify-packager) that would love this magic :smile:

tehasdf:
@max-orlov added them in cloudify-cosmo/cloudify-packager#193

PR #419, opened 2016-06-07
Cfy 5004 upload plugins after bootstrap

geokala:


cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,34 @@ def bootstrap(self,
                 task_retry_interval=task_retry_interval,
                 verbose=verbose,
                 debug=debug).wait()
+            self.upload_plugins(cfy)
```
dankilman: this call doesn't belong here. it should be included as part of the TestEnvironment.bootstrap

cosmo_tester/test_suites/test_simple_manager_blueprint/abstract_single_host_test.py
```diff
@@ -72,8 +72,10 @@ def setup_simple_manager_env(self):
             self.local_env.outputs()['simple_vm_private_ip_address']
 
     def bootstrap_simple_manager_blueprint(self, override_inputs=None):
-        self.manager_blueprints_repo_dir = clone(MANAGER_BLUEPRINTS_REPO_URL,
-                                                 self.workdir)
+        self.manager_blueprints_repo_dir = clone(
+            MANAGER_BLUEPRINTS_REPO_URL, self.workdir,
+            branch=os.environ.get('CLOUDIFY_MANAGER_BLUEPRINTS_BRANCH',
```
dankilman: clone falls back to using `BRANCH_NAME_CORE`, no need for an additional environment variable

cosmo_tester/framework/cfy_helper.py
geokala: It has to be here, as the simple_manager_bootstrap tests call this (because they bootstrap using the simple manager blueprint on another machine).

cosmo_tester/framework/util.py
```diff
@@ -39,6 +40,24 @@
 CLOUDIFY_AUTH_TOKEN_HEADER = 'Authentication-Token'
 
 
+def get_file_name_from_url(url):
+    return url.split('/')[-1]
+
+
+def download_file(url, destination=''):
```
dankilman: it is somewhat surprising for  a general purpose download method to cache results, please remove that if exists part. also, please use requests instead of urllib

cosmo_tester/framework/util.py
```diff
@@ -39,6 +40,24 @@
 CLOUDIFY_AUTH_TOKEN_HEADER = 'Authentication-Token'
 
 
+def get_file_name_from_url(url):
```
dankilman: this util method is only used in one place. please inline

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,34 @@ def bootstrap(self,
...
+
+    def _get_wagons(self, wagons=None):
+        self.logger.info('Downloading Wagons...')
+
+        wagon_paths = []
+        versions_repo_url = 'https://github.com/geokala/' \
```
dankilman: this url should obviously change

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,34 @@ def bootstrap(self,
...
+                            'cloudify-versions.git'
+        path = git_helper.clone(
```
dankilman: `BRANCH_NAME_CORE` is already defined, remove additional environment variable

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,34 @@ def bootstrap(self,
                 task_retry_interval=task_retry_interval,
                 verbose=verbose,
                 debug=debug).wait()
+            self.upload_plugins(cfy)
+
+    def _get_wagons(self, wagons=None):
```
dankilman: this function should probably be named  `_download_wagons`

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,34 @@ def bootstrap(self,
...
+        versions_repo_url = 'https://github.com/geokala/' \
+                            'cloudify-versions.git'
+        path = git_helper.clone(
+            versions_repo_url, self._cfy_workdir,
+            branch=os.environ.get('CLOUDIFY_VERSIONS_BRANCH',
+                                  'master'))
+        url_path = os.path.join(path, 'plugin-urls.yaml')
+        plugins = get_yaml_as_dict(url_path)['plugins']
+        for plugin in plugins:
+            if (plugin['name'] or plugin['wgn_url']) in wagons or not wagons:
```
dankilman: This is line is buggy.
should be
`if not wagons or (plugin['name'] or plugin['wgn_url']) in wagons`
also, why are you testing for both plugin name and plugin wgn_url, should a single value be checked?

cosmo_tester/framework/util.py
dankilman: if you want the caching logic, consider making it optional, with default false

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,34 @@ def bootstrap(self,
...
+                    'Downloading: {0}...'.format(plugin['wgn_url'])
+                )
+                wagon_paths.append(download_file(plugin['wgn_url']))
+        return wagon_paths
+
+    def upload_plugins(self, cfy, chosen_wagons=None):
+        downloaded_wagon_paths = self._get_wagons(chosen_wagons or [])
+        for wagon in downloaded_wagon_paths:
+            self.logger.info('Uploading {0}'.format(wagon))
+            upload = cfy.plugins.upload(p=wagon, verbose=True)
```
dankilman: isn't this call blocking?

cosmo_tester/framework/cfy_helper.py
dankilman: if they are bootstrapping on another machine, how are they calling this to begin with?
shouldn't it be some ssh/winrm call with explicit `cfy bootstrap ...`?

geokala: Just checked the code, and it appears I was mistaken. Actual process of simple manager blueprint test:
1. Create a VM on openstack.
2. Run cfy bootstrap locally using the simple manager blueprint and setting the VM IP as public_ip, which causes the simple manager blueprint to bootstrap on that machine via fabric. It does this using cfy_helper's bootstrap method.

cosmo_tester/framework/cfy_helper.py
geokala: (might need to check with Nir as he wrote this bit, but having just looked in more depth at it):
We're not using the functionality to allow filtering what wagons to install yet, so we could probably remove this and implement it later when it is wanted.

Otherwise, it seems designed to accept a list of plugin names or URLs... but in that case I think actually providing a different location to get the plugin URLs from would be more sensible.

Yeah, I'll remove this for the moment.

cosmo_tester/framework/cfy_helper.py
geokala: It is blocking in one situation: If you run it in an interactive interpreter without assigning its return value, and not in a function.

Otherwise it seems to become non-blocking, which leads to the upload_plugins function returning several minutes before the plugins are actually uploaded.

cosmo_tester/test_suites/test_simple_manager_blueprint/abstract_single_host_test.py
geokala: This functionality isn't actually that useful. I need to get a specific commit from the master repo, but if I set BRANCH_NAME_CORE, it tries to use that for everything. Either way, I'll work around that by hard coding it locally.

cosmo_tester/framework/cfy_helper.py
dankilman: so
1) This call indeed doesn't belong here
2) A fix for those tests is still required

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,37 @@ def bootstrap(self,
...
+
+    def _download_wagons(self):
+        self.logger.info('Downloading Wagons...')
+
+        wagon_paths = []
+
+        plugin_urls_location = os.environ.get(
+            'PLUGIN_URLS_LOCATION',
+            'https://raw.githubusercontent.com/'
+            'cloudify-cosmo/cloudify-versions/master/plugin-urls.yaml',
```
dankilman: do
`'https://raw.githubusercontent.com/cloudify-cosmo/cloudify-versions/{0}/plugin-urls.yaml'.
format(os.env.get(BRANCH_NAME_CORE, 'master'))` instead

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,37 @@ def bootstrap(self,
...
+
+        plugins = yaml.load(
+            requests.get(plugin_urls_location).text
+        )['plugins']
+        for plugin in plugins:
+            self.logger.info(
+                'Downloading: {0}...'.format(plugin['wgn_url'])
+            )
+            wagon_paths.append(
+                download_file(plugin['wgn_url'], keep_existing=True)
```
dankilman: you probably want to specify some destination if you want the caching logic to take place

cosmo_tester/framework/cfy_helper.py
geokala: I can't see where it does belong then, since it is needed on every bootstrap, and every bootstrap uses this. Moving it to the testenv bootstrap method will mean that we have to arbitrarily call the upload_plugins method in some tests, but not others.

dankilman: are there any tests that call cfy_helper bootstrap directly?

geokala: Yes, the simple_manager_bootstrap ones.

dankilman: ok then.

cosmo_tester/framework/cfy_helper.py
geokala: Hmm. Good point. I'll try removing it- can't see why it should cause any problems, and at the moment it's only caching for the duration of one test as I think it dumps into the temporary workdir.

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,39 @@ def bootstrap(self,
...
+        for plugin in plugins:
+            self.logger.info(
+                'Downloading: {0}...'.format(plugin['wgn_url'])
+            )
+            wagon_paths.append(
+                download_file(plugin['wgn_url'])
+            )
+        return wagon_paths
+
+    def upload_plugins(self, cfy):
```
dankilman: no need to pass cfy, it is used globally in this module

cosmo_tester/framework/cfy_helper.py
```diff
@@ -97,6 +101,39 @@ def bootstrap(self,
...
                 debug=debug).wait()
+            self.upload_plugins(cfy)
+
+    def _download_wagons(self):
+        self.logger.info('Downloading Wagons...')
+
+        wagon_paths = []
+
+        plugin_urls_location = os.environ.get(
+            'PLUGIN_URLS_LOCATION',
```
dankilman: i would again prefer it if you didn't add more environment variables

PR #426, opened 2016-06-09
Add support for image bakery tests

funkyHat:
Add cert support to create_rest_client()


funkyHat:
was based on old obsolete test framework. closing

PR #433, opened 2016-06-16
CFY-5317 rollback idempotency

tehasdf:


tehasdf:
Requires #434 for the assertions.

cosmo_tester/test_suites/test_manager_upgrade/rollback_idempotency_test.py
```diff
@@ -0,0 +1,113 @@
...
+        one. Note however that rollback will use the url that is stored
+        on the manager in the rollback properties file, so we need to
+        alter that file.
+        """
+        base_dir = path('/opt/cloudify/sanity')
+        fetched_properties = StringIO()
+        fetched_resources = StringIO()
+
+        properties_path = base_dir / 'node_properties_rollback/properties.json'
+        resources_path = base_dir / 'resources_rollback/__resources.json'
```
nirbir: Why change both properties and rollback resources? When performing rollback the rollback resource is used, so it should be enough to only change that. No?

cosmo_tester/test_suites/test_manager_upgrade/manager_upgrade_base.py
```diff
@@ -187,17 +225,14 @@ def _load_private_ip_from_env(self, workdir):
...
         self.manager_cfy.bootstrap(self.bootstrap_blueprint,
                                    inputs_file=inputs_path)
 
         self.upgrade_manager_ip = self.manager_cfy.get_management_ip()
         self.manager_private_ip = self._load_private_ip_from_env(
-                self.cfy_workdir)
-
-        # TODO: why is this needed?
-        self.manager_cfy.use(management_ip=self.upgrade_manager_ip)
+            self.cfy_workdir)
```
nirbir: why don't we need cfy use anymore?

tehasdf: I'm not sure it was ever needed - we bootstrap using that cfy, so the manager ip/etc are already stored, after bootstrapping.

I think maybe it was only needed because i was using a broken cfy checkout when first writing it? One that didn't store the workdir settings...

cosmo_tester/test_suites/test_manager_upgrade/rollback_idempotency_test.py
tehasdf: Currently, when performing a rollback the property is used, but the resources.json is still checked! So we could only change properties, but then it would fail with a weird error. Just changing the value in resources does nothing - only the key is checked against, the value isn't used at all.

cosmo_tester/test_suites/test_manager_upgrade/rollback_idempotency_test.py
```diff
@@ -0,0 +1,113 @@
...
+        self.prepare_manager()
+        preupgrade_deployment_id = self.deploy_hello_world('pre-')
+
+        self.upgrade_manager()
+        self.post_upgrade_checks(preupgrade_deployment_id)
+
+        try:
+            self.fail_rollback_manager()
+        except sh.ErrorReturnCode as e:
+            error_msg = "fake.tar.gz': no such file or directory"
```
tehasdf: quote

PR #448, opened 2016-07-07
Cfy 5686 use old cli for bootstrap

adamlavie:


cosmo_tester/test_suites/test_manager_upgrade/manager_upgrade_base.py
```diff
@@ -257,17 +260,57 @@ def _load_private_ip_from_env(self, workdir):
...
-            self.manager_inputs, self._testMethodName)
 
-        self.manager_cfy.bootstrap(self.bootstrap_blueprint,
-                                   inputs_file=inputs_path)
+        with self._bootstrap_old_version():
+            self.upgrade_manager_ip = self._load_public_ip_from_env(
+                    os.getcwd())
+            self.manager_private_ip = self._load_private_ip_from_env(
+                    os.getcwd())
+        self.cfy.init(reset_config=True).wait()
```
ran-z: why not use self.manager_cfy?

cosmo_tester/test_suites/test_manager_upgrade/manager_upgrade_base.py
```diff
@@ -257,17 +260,57 @@ def _load_private_ip_from_env(self, workdir):
...
+            if os.path.isdir(tmp_workdir):
+                shutil.rmtree(tmp_workdir, ignore_errors=True)
+
+    def _execute_command(self, command):
+        process = subprocess.Popen(command, stdout=subprocess.PIPE)
+
+        while process.poll() is None:
+            l = process.stdout.readline()
+            print l
+        print process.stdout.read()
```
ran-z: ?
was this for debugging purposes or am i missing something? :)

cosmo_tester/test_suites/test_manager_upgrade/manager_upgrade_base.py
```diff
@@ -37,7 +38,9 @@
 
 BOOTSTRAP_REPO_URL = 'https://github.com/cloudify-cosmo/' \
                      'cloudify-manager-blueprints.git'
+
 BOOTSTRAP_BRANCH = '3.4rc1'
+BOOTSTRAP_CLI_VENV_PATH = '/tmp/old_cli_env'
```
ran-z: why have this explicit?
use tempfile's make temp dir method

cosmo_tester/test_suites/test_manager_upgrade/manager_upgrade_base.py
```diff
@@ -257,17 +260,57 @@ def _load_private_ip_from_env(self, workdir):
...
+            self.manager_private_ip = self._load_private_ip_from_env(
+                    os.getcwd())
+        self.cfy.init(reset_config=True).wait()
+        self.cfy.use(self.upgrade_manager_ip).wait()
 
-        self.upgrade_manager_ip = self.manager_cfy.get_management_ip()
-        self.manager_private_ip = self._load_private_ip_from_env(
-            self.cfy_workdir)
+    @contextmanager
+    def _bootstrap_old_version(self):
```
ran-z: i dont really like this use of a contextmanager - i think there's too much logic happening here for it to be a reasonable use. there's a whole bootstrap going on here :D
I'd have one method which bootstraps old version (up to "try" part), another which cleans up (the finally part), and just call them with a try-finally clause from the outside.
if there were more places calling this and requiring the same logic maybe it'd make sense doing it this way but thats not the case here

PR #468, opened 2016-08-09
fixing input settins in broker test

noakup:


cosmo_tester/test_suites/test_broker_security/broker_security_test_base.py
```diff
@@ -72,6 +72,11 @@ def _update_manager_blueprint(self):
             for key, value in props.items():
                 patch.set_value(key, value)
 
+        props = self.get_manager_blueprint_inputs_override()
```
dankilman: you are calling `get_manager_blueprint_inputs_override` which is only defined in one subclass.
This will raise `AttributeError` when used by other subclasses

PR #497, opened 2017-01-10
CFY-6286-use-correct-manager_resources_package

dusking:


cosmo_tester/framework/util.py
```diff
@@ -219,12 +221,13 @@ def get_auth_header(username=None, password=None, token=None):
 def create_rest_client(manager_ip,
                        manager_username=None,
                        manager_password=None):
+    headers = get_auth_header(
```
idanmo: Wouldn't it be better to change the get_auth_header function to also accept tenant and return the complete header?

cosmo_tester/framework/testenv.py
```diff
@@ -128,6 +130,12 @@ def __init__(self):
             self.handler_configuration = self.suites_yaml[
                 'handler_configurations'][handler_configuration]
 
+        variables = self.suites_yaml['variables']
+        premium_resources_package = variables.get('premium_resources_package')
```
idanmo: Can't we cal the variable: "manager_resources_package_url" ?

cosmo_tester/framework/util.py
dusking: Sure, we can, but tenant is not part of the auth.. 
We can create another function that will return the header, but it seems too much..

idanmo: Yes but now tenant in auth is mandatory isn't it?
What about community edition?

dusking: Tenant is mandatory for every api. But the auth mechanism is based on user & password, not tenant.. It's like a required parameter..  
Community edition is using the default_tenant.

PR #512, opened 2017-03-23
Fix autoheal test

geokala:


cosmo_tester/test_suites/test_blueprints/nodecellar_test_autoheal.py
```diff
@@ -79,7 +79,7 @@ def _test_nodecellar_impl(self, blueprint_file):
 
         self.post_autoheal_assertions(after_install, after_autoheal)
 
-        self.execute_uninstall()
+        self.uninstall_delete_deployment_and_blueprint()
```
geokala: This brings it in line with the basic nodecellar test and means we don't try to run an uninstall twice on the same deployment, which now causes errors.

PR #525, opened 2017-05-07
CFY-6743 HA system tests

dmitryAII:


idanmo:
I think it would be better to put all the tests in an `ha` folder next to `snapshots`: `cosmo_tester/test_suites/ha`

cosmo_tester/resources/terraform/openstack-vm.tf.template
```diff
@@ -75,6 +75,36 @@ resource "openstack_compute_secgroup_v2" "security_group" {
     ip_protocol = "tcp"
     cidr = "${openstack_networking_subnet_v2.subnet.cidr}"
   }
```
dmitryAII: Yes

cosmo_tester/test_suites/image_based_tests/ha/ha_negative_test.py
```diff
@@ -0,0 +1,171 @@
...
+        cfy('--version')
+
+
+def test_remove_from_cluster_and_rejoin_negative(cfy,
+                                                 cluster, logger):
+    manager1 = cluster.managers[0]
+    manager2 = cluster.managers[1]
+
+    manager1.use()
+    logger.info('Removing the standby manager {0}'.format(manager2.ip_address)
```
dmitryAII: Done

cosmo_tester/test_suites/image_based_tests/ha/ha_negative_test.py
```diff
@@ -0,0 +1,171 @@
...
+    cfy.cluster.nodes.list()
+
+    logger.info('Trying to rejoin HA cluster with a manager previously'
+                ' removed from cluster')
+    with pytest.raises(Exception) as exinfo:
+        cfy.cluster.join(manager1.ip_address,
+                         timeout=600,
+                         cluster_host_ip=manager2.ip_address,
+                         cluster_node_name=manager2.index)
+    assert 'is already part of a Cloudify Manager cluster' \
```
dmitryAII: Agreed with you that no fix is needed

cosmo_tester/resources/terraform/openstack-vm.tf.template
tehasdf: since the test code passes `cluster_host_ip` to `.start()`/`.join()` calls, and the managers are in the same subnet, it should work without those rules - but we need to pass `cluster_host_ip=manager1.private_ip_address`, not `.ip_address`

cosmo_tester/test_suites/ha/ha_cluster_of_three_test.py
```diff
@@ -0,0 +1,207 @@
...
+    try:
+        encryption_key = ha_helper.create_encription_key()
+        manager1 = cluster.managers[0]
+        manager2 = cluster.managers[1]
+        manager3 = cluster.managers[2]
+
+        manager1.use()
+        cfy.cluster.start(timeout=600,
+                          cluster_host_ip=manager1.ip_address,
+                          cluster_node_name='m1',
```
tehasdf: please note `ha_helper.verify_nodes_status` expects the node name to be manager.index so this will currently break

cosmo_tester/resources/terraform/openstack-vm.tf.template
dmitryAII: We tested and without those rules it didn't work. But it was about a month ago. I spoke with Omer about it and he said to me that this is a bug in design.  Something changed?

tehasdf: can't think of anything that changed; there was some work on the CLI about splitting "profile name" and "manager ip", so perhaps it was broken before; seems to work fine for me right now without those rules, can you try and see please?

cosmo_tester/test_suites/image_based_tests/ha/ha_cluster_of_two_test.py
```diff
@@ -0,0 +1,215 @@
+########
```
dmitryAII: Done

PR #526, opened 2017-05-07
CFY-6794 fix bootstrap in system tests

yarivgiga:


cosmo_tester/framework/util.py
```diff
@@ -322,14 +322,28 @@ def get_cli_package_urls():
...
+    else:
+        package_url_file = Path(
+                os.path.abspath(os.path.join(
+                        os.path.dirname(cosmo_tester.__file__),
+                        '../..',
+                        MANAGER_PACKAGE_URL_FILE)))
+        if not package_url_file.exists():
+            raise IOError('File containing manager premium single tar URL not '
+                          'found: {}'.format(package_url_file))
     return package_url_file.text().strip(os.linesep)
```
idanmo: Just indent the last return statement as it is only relevant for the else part.

PR #533, opened 2017-05-20
CFY-6900 test snapshot upgrades

funkyHat:


cosmo_tester/framework/cluster.py
```diff
@@ -80,12 +116,13 @@ def __str__(self):
         return 'Cloudify manager [{}:{}]'.format(self.index, self.ip_address)
 
     @retrying.retry(stop_max_attempt_number=3, wait_fixed=3000)
-    def use(self):
-        self._cfy.profiles.use('{0} -u {1} -p {2} -t {3}'.format(
-                self.ip_address,
-                self._attributes.cloudify_username,
-                self._attributes.cloudify_password,
-                self._attributes.cloudify_tenant).split())
+    def use(self, tenant=None):
```
idanmo: why?

cosmo_tester/framework/cluster.py
```diff
@@ -121,17 +158,114 @@ def verify_services_are_running(self):
                     'service {0} is in {1} state'.format(
                             service['display_name'], instance['SubState'])
 
+    @retrying.retry(stop_max_attempt_number=10, wait_fixed=5000)
+    def assert_snapshot_created(self, snapshot_id, attributes):
```
idanmo: This function should probably be moved to `util`.

cosmo_tester/framework/cluster.py
```diff
@@ -327,7 +451,7 @@ def _upload_plugin_to_manager(self, manager, plugin_name):
             x.id for x in manager.client.plugins.list()]
         try:
             manager.client.plugins.upload(plugin_wagon[0])
-            self._cfy.plugins.list()
+            manager.client.plugins.list()
```
idanmo: Why? the idea here was to print the plugins list using the `cfy` command..

cosmo_tester/resources/attributes.yaml
```diff
@@ -4,8 +4,10 @@ cloudify_username: admin
 cloudify_password: admin
 cloudify_tenant: default_tenant
 
-cloudify_manager_4_0_image_name: cloudify-manager-premium-4.0
 cloudify_manager_image_name_prefix: cloudify-manager-premium
+cloudify_manager_4_0_1_image_name: cloudify-manager-premium-4.0.1
+cloudify_manager_4_0_image_name: cloudify-manager-premium-4.0
+cloudify_manager_3_4_image_name: manager3.4.2-insecure-image
```
idanmo: Perhaps call this key: `cloudify_manager_3_4_2_image_name` ?

cosmo_tester/framework/cluster.py
funkyHat: Because when restoring from 3.4 we have to specify a new tenant to restore into, & then we will need to connect to it to check the deployment later

cosmo_tester/framework/cluster.py
funkyHat: `cfy` isn't backwards compatible so it breaks when trying to run on the 3.4.2 manager.
If it were practical I'd prefer that we didn't use `cfy` at all in these tests (even if it's decided that we will fix backwards comaptibility of cfy/rest-client). It's an extra variable which could be broken for other reasons.

cosmo_tester/framework/cluster.py
funkyHat: Yep. I think I assumed it might need different implementations per manager version, but that hasn't been the case. We can always move it back if that happens with some future version...

PR #534, opened 2017-05-20
Add cloudify-agent dev dependency

funkyHat:


idanmo:
This is not going to work as the fabric version from agent will override the one required by cli.

funkyHat:
cli & agent fabric deps have been fixed, so this can now be merged

PR #536, opened 2017-06-11
CFY-6915 Inplace upgrade system test

yarivgiga:


cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+from cosmo_tester.framework.cluster import CloudifyCluster
+from cosmo_tester.framework.examples.hello_world import HelloWorldExample
+
+
+cluster = None
+
+
+@pytest.fixture(scope='module')
+def manager(request, cfy, ssh_key, module_tmpdir, attributes, logger):
```
funkyHat: I think you should just be able to use `cosmo_tester.framework.fixtures. bootstrap_based_manager` here rather than defining a new fixture

yarivgiga: The difference is that i save the cluster in a global var.
Need to use it later for the second bootstrap.
I added a comment about that.

funkyHat: Aha I see.
The second `global` is not needed though (line 68) as you're not reassigning `cluster`, I think exposing the cluster directly from your fixture instead of the manager would be neater than using globals

yarivgiga: Thanks, that's actually a nice touch.
I'll give it a shot.

PR #539, opened 2017-07-04
CFY-7054 Fix snapshot restore test for 3.4.2 snapshots

mcouthon:


cosmo_tester/test_suites/snapshots/snapshot_upgrade_test.py
```diff
@@ -126,43 +134,31 @@ def test_restore_snapshot_and_agents_upgrade(
...
-    manager2.use(tenant=manager1.restore_tenant_name)
-    client = create_rest_client(
-        manager2.ip_address,
-        username=cluster._attributes.cloudify_username,
-        password=cluster._attributes.cloudify_password,
-        tenant=manager1.tenant_name,
-        api_version=manager2.api_version,
-        )
-
     cfy.deployments.list()
```
geokala: This won't do what it should be doing (listing the deployments for diagnostic purposes) unless the manager2.use is put back in.

geokala: No, no, this is not right, don't worry- you're right, you still have the use there, I missed it in the diff!

cosmo_tester/test_suites/snapshots/snapshot_upgrade_test.py
```diff
@@ -100,24 +98,34 @@ def _hello_world_example(cluster, attributes, logger, tmpdir):
...
     local_snapshot_path = str(tmpdir / 'snapshot.zip')
 
     logger.info('Downloading snapshot from old manager..')
-    manager1.client.snapshots.list()
-    manager1.client.snapshots.download(snapshot_id, local_snapshot_path)
+    manager0.client.snapshots.list()
+    manager0.client.snapshots.download(snapshot_id, local_snapshot_path)
+
+    manager1.use()
+    if manager0.branch_name == '3.4.2':
```
idanmo: Perhaps this should be: if '3.4' in manager0.branch_name?

PR #543, opened 2017-07-18
CFY-6966 Test snapshots

geokala:


geokala:
There might be a bit more prettification that can be done, and this doesn't check whether diamond still sends data after the restore (but... that should probably be covered by the diamond plugin?).

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -62,58 +278,260 @@ def wait_for_execution(client, execution, logger):
...
+                name=deployment['id'],
+            )
+            logger.info(
+                'Checking deployment env for {name} was recreated.'.format(
+                    name=deployment['id'],
+                )
+            )
+            fabric_ssh.run('test -d {path}'.format(path=path))
+            logger.info('Deployment environment was recreated.')
+    if tenant:
```
mcouthon: Ok, after enough of those, I think there's a place for a func like:

```
  def log_with_tenant(msg, log_type=INFO, tenant=None):
      if log_type==INFO:
          log = logger.info
      elif ...
    
      if tenant:
           msg += ' for tenant {0}'.format(tenant)

      log(msg)
```

If you disagree, that's fine, but consider it - there are _12_ `if tenant` statements here.

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -62,58 +278,260 @@ def wait_for_execution(client, execution, logger):
...
+    cluster = CloudifyCluster.create_image_based(
+            cfy,
+            ssh_key,
+            module_tmpdir,
+            attributes,
+            logger,
+            managers=managers,
+            )
+
+    if request.param == '4.0.1':
```
mcouthon: Can this reside somewhere in `cluster.py`, where the manager is created? Maybe add a `run_after_creation` interface? I don't like it here.

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -62,58 +278,260 @@ def wait_for_execution(client, execution, logger):
...
+                'wget http://repository.cloudifysource.org/cloudify/4.0.1/patch3/cloudify-401-te-patch-3.tar.gz && '  # noqa
+                'tar --strip-components=1 -xzf *.tar.gz && '
+                './apply-patch.sh'
+            )
+
+    # gcc and python-devel are needed to build most of our infrastructure
+    # plugins.
+    # As we need to test from source installation of plugins, we must have
+    # these packages installed.
+    for manager in managers[:2]:
```
mcouthon: Explain the list comprehension at `managers[:2]`. Why only the first two?

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
-from cosmo_tester.framework.util import assert_snapshot_created
-
-from cloudify_cli.constants import DEFAULT_TENANT_NAME
-
 from . import (
-    deployment_id,
-    deploy_helloworld,
-    wait_for_execution,
-    update_client_tenant
+    assert_hello_worlds,
```
mcouthon: I don't have anything concrete to say here, except that when I see a list this long, it makes me think that there's something wrong with the design. I would assume that only the high level functions would be necessary here, and the rest would be internal to the implementation.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
-            attributes,
-            logger,
-            managers=managers,
-            )
 
+def test_restore_snapshot_and_agents_upgrade_multitenant(
+        cfy, cluster_multitenant, attributes, logger, tmpdir):
+    local_snapshot_path = str(tmpdir / 'snapshot.zip')
+    new_tenants = ('tenant1', 'tenant2')
+    hello_tenants = ('default_tenant', new_tenants[0])
```
mcouthon: Explain what hello tenants are and what noinstall tenants are

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
+    noinstall_tenants = new_tenants
+
+    old_manager = cluster_multitenant.managers[0]
+    new_manager = cluster_multitenant.managers[1]
+    hello_vms = cluster_multitenant.managers[2:]
+
+    hello_vm_mappings = {
+        hello_tenants[0]: cluster_multitenant.managers[2],
```
mcouthon: Again, this comes back to the fact that these VMs are managers. This is really awkward and confusing.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
 from . import (
-    deployment_id,
-    deploy_helloworld,
-    wait_for_execution,
-    update_client_tenant
+    assert_hello_worlds,
+    check_deployments,
+    check_from_source_plugin,
+    check_plugins,
+    _cluster,
```
mcouthon: Why the internal func?

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -24,8 +33,204 @@
...
+    assert_hello_worlds(hello_vms, installed=False, logger=logger)
+
+
+def delete_manager(manager, logger):
+    logger.info('Deleting {version} manager..'.format(
+        version=manager.branch_name))
+    manager.delete()
+
+
+def create_noinstall_helloworld(manager, logger, tenant=None):
```
mcouthon: Let's maybe call it something different. It's a really confusing name, especially considering what it actually does.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
-    cfy.snapshots.list()
-
-    logger.info('Restoring snapshot on latest manager..')
-    restore_execution = manager1.client.snapshots.restore(snapshot_id)
-    logger.info('Snapshot restore execution:%s%s',
-                os.linesep,
-                json.dumps(restore_execution, indent=2))
-
-    cfy.executions.list(['--include-system-workflows'])
+def create_tenant_secrets(manager, tenants, logger):
```
mcouthon: Add an explanation for `create_tenant_secrets ` and `check_tenant_secrets `. Why is it here, what does it do, etc.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
+            tenants=', '.join(tenants),
+        )
+    )
+    for tenant in tenants:
+        check_from_source_plugin(manager, plugin, deployment_id, logger,
+                                 tenant)
+    logger.info('Plugins installed from source were installed correctly.')
+
+
+def make_tenants(manager, logger, tenants=('tenant1', 'tenant2')):
```
mcouthon: Nitpicking again - but why not `create_tenants`?

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
+        upload_and_install_helloworld(attributes, logger, old_manager,
+                                      hello_vm_mappings[tenant],
+                                      tmpdir, tenant=tenant, prefix=tenant)
+    for tenant in noinstall_tenants:
+        create_noinstall_helloworld(old_manager, logger, tenant=tenant)
+
+    create_tenant_secrets(old_manager, tenants, logger)
+
+    old_plugins = {
+        tenant: upload_test_plugin(old_manager, logger, tenant)
```
mcouthon: I'd rather we separate it into:
```
for tenant in tenants:
    upload_test_plugin(..._

old_plugins = {tenant: list_plugins(...) for tenant in tenants}
```

Just so it's more readable and consistent.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -13,141 +13,182 @@
...
+        for tenant in tenants
+    }
+
+    create_snapshot(old_manager, SNAPSHOT_ID, attributes, logger)
+    download_snapshot(old_manager, local_snapshot_path, SNAPSHOT_ID, logger)
+    upload_snapshot(new_manager, local_snapshot_path, SNAPSHOT_ID, logger)
+
+    restore_snapshot(new_manager, SNAPSHOT_ID, cfy, logger)
+
+    assert_hello_worlds(hello_vms, installed=True, logger=logger)
```
mcouthon: Should we also (only?) check this after upgrading agents?

cosmo_tester/test_suites/snapshots/single_tenant_snapshot_test.py
```diff
@@ -0,0 +1,145 @@
...
+    # The old manager needs to exist until the agents install is run
+    delete_manager(old_manager, logger)
+
+    remove_and_check_deployments([hello_vm], new_manager, logger)
+
+
+@pytest.fixture(
+        scope='module',
+        params=['4.0', '3.4.2'])
+def cluster_singletenant(request, cfy, ssh_key, module_tmpdir, attributes,
```
mcouthon: Just to be consistent - I think in most of the tests the fixtures are above the test functions

geokala: They're gradually becoming not that way wherever I'm changing them, so that when one first looks at the file one sees the tests rather than a nest of random fixtures, functions, and other garbage with tests nestled in the middle.

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -62,58 +278,260 @@ def wait_for_execution(client, execution, logger):
...
-        deployment_id,
-        'install',
-        )
-    logger.info('waiting for installation to finish')
-    wait_for_execution(
-        manager.client,
-        execution,
-        logger,
-        )
+def create_snapshot(manager, SNAPSHOT_ID, attributes, logger):
```
geokala: Whoops, I think I copied that from a call that was just using the constant.

cosmo_tester/test_suites/snapshots/__init__.py
geokala: Is this something that we feel should be applied for all tests? If I move it to cluster.py then it's more likely to end up included in other tests, so we should probably be sure whether we want that or not. If it's not intended to appear in any other tests then I think it probably belongs here?

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
geokala: The problem is that the single and multi tenants both make use of these but with differences that I think would overcomplicate any general high level method shared between them- it would of course be possible to generalise it, but I think the complexity of the shared method would outweigh the gains.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
geokala: It's internal to the snapshot tests- they were originally all in one file. I could un-internal it... and in fact I probably will, given that so much else is effectively internal here but these are tests so... yeah, probably best to just uninternalise them.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
geokala: I'm adding a check afterwards as well. It's a fairly quick check (except when it fails, though even then I think we're talking 30 seconds at most).

That'll mean that it'll allow narrowing down the failure to before or after the agent install if there is strangeness.

cosmo_tester/test_suites/snapshots/__init__.py
mcouthon: I think that it's the right thing to refer to the version of the code that the customers have, and they will almost definitely have the patch.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
mcouthon: Ok, fair enough.

cosmo_tester/test_suites/snapshots/single_tenant_snapshot_test.py
mcouthon: I'm not sure I agree. While function names are easily traceable in IDEs, fixtures aren't - they "magically" appear in the function definition, and then it's unclear where they're coming from. It is, of course, debatable.

geokala: I agree on that point- fixtures aggravate me due to that.

I wonder if it would be sensible for us to have a standard for naming fixtures so it is obvious that they are fixtures?

tehasdf: it is somewhat obvious that they *are* fixtures, maybe less so where they're defined (but so far i think it's always either conftest.py, or the same file as the test; if there are exceptions to that, then perhaps there shouldn't be)
if you mean something like naming them `cfy_fixture` then please don't do that, that's ugly, and doesn't solve any problems :(

PR #547, opened 2017-07-26
CFY-7098, CFY-7099, CFY-7114 Test improvements

geokala:
Removes obsolete tests and allows specifying a manager image name in the attributes.

Also allows setting a community flag which will conditionally skip some tests (only HA at the moment) if it is set.

CFY-7099 is also intended to modify snapshot tests behaviour, but that requires the merging of https://github.com/cloudify-cosmo/cloudify-system-tests/pull/543 which is waiting on the manager build.


cosmo_tester/test_suites/ha/__init__.py
```diff
@@ -0,0 +1,5 @@
+import pytest
```
mcouthon: Won't we need to use this in some other places beside HA?

PR #560, opened 2017-08-10
CFY-7098 Allow selecting latest manager image

geokala:


cosmo_tester/framework/cluster.py
```diff
@@ -275,15 +275,24 @@ def _get_latest_manager_image_name():
...
     Returns: "cloudify-manager-premium-4.0m15"
     """
-    version = util.get_cli_version()
-    version_num, _, version_milestone = version.partition('-')
+    specific_manager_name = ATTRIBUTES.cloudify_manager_latest_image.strip()
 
-    if version_num.endswith('.0') and version_num.count('.') > 1:
-        version_num = version_num[:-2]
+    if specific_manager_name:
+        image_name = specific_manager_name
```
tehasdf: if you just `return specific_manager_name` here, you can just dedent the rest of the function, reducing diff line count (and making this nicer overall)

cosmo_tester/resources/attributes.yaml
```diff
@@ -4,6 +4,10 @@ cloudify_username: admin
 cloudify_password: admin
 cloudify_tenant: default_tenant
 
+# If the latest image name is not set, the latest image name will be generated
+# based on the image name prefix and the currently installed CLI version
+# e.g. cloudify-manager-premium-4.2.dev1
+cloudify_manager_latest_image: ''
```
tehasdf: does the "latest" name really fit? it's not ever used to mean latest, is it

geokala: It... kinda isn't (it's the one for the currently installed CLI), but the name of the function was _get_latest(etc) so I was maintaining that.

Do you think it's worth doing something about the naming here or is it worth adding comments to https://cloudifysource.atlassian.net/browse/CFY-7159 ?

cosmo_tester/framework/cluster.py
geokala: I was actually trying to follow the 'only one return' approach, partially for readability (I think it makes it clearer when you can follow one branch or the other- immediately visually obvious that the other branch only gets executed when <condition> is false), and partially because if we need to do any post-processing on the image later (or logging, e.g. of image UUID) then it becomes easier to.

Of course, the second point may never be needed.

How strongly do you feel either way about it? I don't think we have a strong style preference on this in the code base?

tehasdf: i do think that in this func, multiple returns would be cleaner, as you'd just return early, and could forget about the "specific_image" part altogether, while reading the rest of the function. Which means there's one less thing to keep in your head while reading the harder part.

But i don't really care that much, it's just a small style nitpick.
And of course, if we ever need to do any post-processing, then sure.

cosmo_tester/resources/attributes.yaml
tehasdf: I don't have any idea for a better name, so if you don't either, let's go with this one for now :)

cosmo_tester/framework/cluster.py
geokala: OK, I'll keep it in mind for CFY-7159 when I'll be looking in more detail at cluster.py anyway.

PR #561, opened 2017-08-17
Fail fast on bootstrap tests without fabric

geokala:


geokala:
Noted here for posterity:
bootstrap_based_tests isn't imported anywhere (according to grep), and a run of pytest --collect-only without the fabric plugin installed lists everything except the bootstrap tests, then fails on each of the bootstrap test files with a complaint about fabric plugin needing to be installed, so all looks good.

PR #562, opened 2017-08-19
CFY-7099 Support community images for testing

geokala:
I've run all of the tests with the default attributes.yaml (testing on the default latest manager image for all tests on premium) and have had them all work, so there should be no negative impact on currently running tests.

The community tests have the following issues (which I don't think are down to problems with this PR):
1. The snapshot restore tests fail on 4.1 -> current, because of the presence of the cloudify stage files which cause the community manager to try to perform a restore of stage. This is probably best fixed by specifying community versions of the older managers too.
2. Bootstrap tests are currently failing because the bootstrap fails. I believe this is a problem with the current resources single tar.

### DIFF
--- a/cosmo_tester/framework/cluster.py
+++ b/cosmo_tester/framework/cluster.py
@@ -213,9 +213,13 @@ class _CloudifyManager(object):
         status = self.client.manager.get_status()
         for service in status['services']:
             for instance in service['instances']:
-                assert instance['SubState'] == 'running', \
-                    'service {0} is in {1} state'.format(
-                            service['display_name'], instance['SubState'])
+                if (
+                    instance['Id'] == 'cloudify-stage.service'
+                    and not util.is_community()
+                ):
+                    assert instance['SubState'] == 'running', \
+                        'service {0} is in {1} state'.format(
+                                service['display_name'], instance['SubState'])
 
     @abstractproperty
     def branch_name(Self):
@@ -635,14 +639,16 @@ class BootstrapBasedCloudifyCluster(CloudifyCluster):
     def _create_inputs_file(self):
         self._inputs_file = self._tmpdir / 'inputs.json'
         bootstrap_inputs = json.dumps({
-            'public_ip': self.managers[0].ip_address,
-            'private_ip': self.managers[0].private_ip_address,
-            'ssh_user': self._attributes.centos7_username,
-            'ssh_key_filename': self._ssh_key.private_key_path,
-            'admin_username': self._attributes.cloudify_username,
-            'admin_password': self._attributes.cloudify_password,
-            'manager_resources_package': self._manager_resources_package},
-            indent=2)
+                'public_ip': self.managers[0].ip_address,
+                'private_ip': self.managers[0].private_ip_address,
+                'ssh_user': self._attributes.centos7_username,
+                'ssh_key_filename': self._ssh_key.private_key_path,
+                'admin_username': self._attributes.cloudify_username,
+                'admin_password': self._attributes.cloudify_password,
+                'manager_resources_package': self._manager_resources_package,
+            },
+            indent=2,
+        )
         self._logger.info(
                 'Bootstrap inputs:%s%s', os.linesep, bootstrap_inputs)
         self._inputs_file.write_text(bootstrap_inputs)
--- a/cosmo_tester/resources/attributes.yaml
+++ b/cosmo_tester/resources/attributes.yaml
@@ -4,6 +4,12 @@ cloudify_username: admin
 cloudify_password: admin
 cloudify_tenant: default_tenant
 
+# This should be set to true if testing community images.
+# This isn't detected by the framework, as doing so would require either some
+# heuristics or deploying managers for each test before determining whether it
+# could be skipped, both of which have a good chance of wasting a lot of time.
+image_is_community: false
+
 # If the latest image name is not set, the latest image name will be generated
 # based on the image name prefix and the currently installed CLI version
 # e.g. cloudify-manager-premium-4.2.dev1
--- a/cosmo_tester/test_suites/ha/__init__.py
+++ b/cosmo_tester/test_suites/ha/__init__.py
@@ -1,0 +1,5 @@
+import pytest
+from cosmo_tester.framework.util import is_community
+
+skip_community = pytest.mark.skipif(is_community(),
+                                    reason='Community does not support HA')
--- a/cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
+++ b/cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
@@ -19,6 +19,12 @@ import time
 from cosmo_tester.framework.examples.hello_world import HelloWorldExample
 from cosmo_tester.framework.cluster import CloudifyCluster
 from .ha_helper import HighAvailabilityHelper as ha_helper
+from . import skip_community
+
+
+# Skip all tests in this module if we're running community tests,
+# using the pytestmark magic variable name
+pytestmark = skip_community
 
 
 @pytest.fixture(scope='function', params=[2, 3])
--- a/cosmo_tester/test_suites/snapshots/__init__.py
+++ b/cosmo_tester/test_suites/snapshots/__init__.py
@@ -24,6 +24,7 @@ from cosmo_tester.framework.cluster import (
 )
 from cosmo_tester.framework.util import (
     assert_snapshot_created,
+    is_community,
 )
 # CFY-6912
 from cloudify_cli.commands.executions import (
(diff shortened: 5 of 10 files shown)


cosmo_tester/test_suites/ha/__init__.py
```diff
@@ -0,0 +1,5 @@
+import pytest
+from cosmo_tester.framework.util import is_community
+
+community_skip = pytest.mark.skipif(is_community(),
```
tehasdf: `is_community` (so, `get_attributes`) is fine to call at import time, right?

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -55,6 +56,33 @@
     TENANT_DEPLOYMENTS_PATH + '/{name}'
 )
 
+MANAGER_VERSIONS_LIST = (
```
tehasdf: sure it's a `_LIST`? looks like a tuple to me ;)))

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -55,6 +56,33 @@
...
+    '3.4.2',
+    '4.0',
+    '4.0.1',
+    '4.1',
+    'master'
+)
+# Only test 3.4.2 and 4.0 with a single tenant on premium.
+# Everything else should work with multi tenant
+MANAGER_SINGLE_TENANT_SLICE = 2
```
tehasdf: this slicing looks pretty fragile to me, will we need to later keep this in sync if the tuple/list changes? Probably not, since future versions SHOULD work with multi tenant of course... but eh, slicing seems just like a weird way to split the list in two.
Why not just have two separate versions lists, `SINGLE_TENANT_VERSIONS` and `MULTI_TENANT_VERSIONS` or something like that. (and use `+` for community-singletenant)

Bonus points if the list doesn't look too general - we probably don't want someone in the future this here is the official list of manager versions, and importing it into some other (non-snapshots) test...

cosmo_tester/test_suites/snapshots/single_tenant_snapshot_test.py
```diff
@@ -67,7 +68,9 @@ def test_restore_snapshot_and_agents_upgrade_singletenant(
 
     assert_hello_worlds([hello_vm], installed=True, logger=logger)
 
-    check_secrets_converted(new_manager, logger)
+    if old_manager.branch_name in ('3.4.2', '4.0'):
+        # Check we convert secrets for pre-service-user versions
```
tehasdf: can you also write "the why" in this comment please? i'm not sure, do we not support converting secrets in singletenant post-4.0?

cosmo_tester/framework/cluster.py
```diff
@@ -213,9 +213,13 @@ def verify_services_are_running(self):
         status = self.client.manager.get_status()
         for service in status['services']:
             for instance in service['instances']:
-                assert instance['SubState'] == 'running', \
-                    'service {0} is in {1} state'.format(
-                            service['display_name'], instance['SubState'])
+                if (
+                    service['display_name'] == 'Cloudify Stage'
```
geokala: That's only true if the ID also remains the same- it feels likely that if we change the display name we'll also change the ID, because it'll probably be the case that we're not calling it 'Stage' any more. It'd be nice if the status didn't display premium services on community managers... worth a Jira?

cosmo_tester/test_suites/ha/__init__.py
geokala: Correct. It just reads a file on the local filesystem.

cosmo_tester/test_suites/snapshots/__init__.py
geokala: Semantically it is a list of manager versions.
Pedantically, it is a tuple, but does contain a (non-python sense) list of manager versions. <smiles>

cosmo_tester/test_suites/snapshots/__init__.py
geokala: Hmm. I think the only change that might happen would be to reduce it by 1 (to make it only be 3.4.2), if we were to patch in support for restoring multi tenant snapshots for 4... but I don't think that's likely to happen.

So I think the only changes absolutely needed in future will be to add in new old manager versions (e.g. adding 4.2 before 'master' when we start working on 4.2.1 or 4.3).

I could tweak it to have SINGLE_TENANT and MULTI_TENANT lists, yes, and probably automatically add 'master' to avoid that needing to be messed with in future. I'm not sure there's any way I can fully discourage someone from copying it as an official list of managers (except that it doesn't include the most recent)... will think about it and tweak this bit on Monday AM (hopefully that'll give time for other comments too).

cosmo_tester/test_suites/snapshots/single_tenant_snapshot_test.py
geokala: We don't, no. I think we did briefly but... I'm fairly sure that was changed (leastways I think I was told it was changed, and it might well have been since after 4.0 everything is using service users so it's harder to make things work with keys in arbitrary locations (though I think we should probably still keep on converting them for safety anyway, but... eh...)

I'll enhance the comment on Monday AM.

cosmo_tester/framework/cluster.py
tehasdf: Hey, reminding users that they're missing out on premium features is a marketing thing! ;P

I hope we'll never change the id, as it's used in many many places (since it's the unit name in systemd). However, this is the only place that depends on the display name, hence my comment.

cosmo_tester/framework/cluster.py
```diff
@@ -609,8 +613,11 @@ class BootstrapBasedCloudifyCluster(CloudifyCluster):
 
     def __init__(self, *args, **kwargs):
         super(BootstrapBasedCloudifyCluster, self).__init__(*args, **kwargs)
-        self._manager_resources_package = \
-            util.get_manager_resources_package_url()
+        if util.is_community():
+            self._manager_resources_package = None
```
mcouthon: It's unclear (to me, at least) why does the fact that it's community implies `None` here. Maybe leave a comment here.

cosmo_tester/framework/cluster.py
```diff
@@ -634,15 +641,18 @@ def _clone_manager_blueprints(self):
...
             'public_ip': self.managers[0].ip_address,
             'private_ip': self.managers[0].private_ip_address,
             'ssh_user': self._attributes.centos7_username,
             'ssh_key_filename': self._ssh_key.private_key_path,
             'admin_username': self._attributes.cloudify_username,
             'admin_password': self._attributes.cloudify_password,
-            'manager_resources_package': self._manager_resources_package},
-            indent=2)
+        }
+        if self._manager_resources_package:
```
mcouthon: What would happen if it's left as `None` here?

cosmo_tester/resources/attributes.yaml
```diff
@@ -4,6 +4,12 @@ cloudify_username: admin
 cloudify_password: admin
 cloudify_tenant: default_tenant
 
+# This should be set to true if testing community images.
+# This isn't detected by the framework, as doing so would require either some
+# heuristics or deploying managers for each test before determining whether it
+# could be skipped, both of which have a good chance of wasting a lot of time.
+image_is_community: false
```
mcouthon: What if we want to test via Jenkins? Shouldn't it be possible to set this via something else, like an env var, from the Jenkins?

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
```diff
@@ -19,6 +19,7 @@
 from cosmo_tester.framework.examples.hello_world import HelloWorldExample
 from cosmo_tester.framework.cluster import CloudifyCluster
 from .ha_helper import HighAvailabilityHelper as ha_helper
+from . import community_skip
```
mcouthon: Sorry for nitpicking, but why not `skip_community`?

mcouthon: Also, isn't there a way in pytest to skip a whole module? I feel like there's got to be.

cosmo_tester/test_suites/snapshots/__init__.py
mcouthon: I agree with Lukasz. Having two lists feels more readable, especially since we have `get_single_tenant_versions_list` and `get_multi_tenant_versions_list` anyway

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
```diff
@@ -123,7 +124,7 @@ def test_restore_snapshot_and_agents_upgrade_multitenant(
 
 @pytest.fixture(
         scope='module',
-        params=['master', '4.1', '4.0.1'])
+        params=get_multi_tenant_versions_list())
```
mcouthon: Yeah, I like that. No more keeping 17 different version lists

cosmo_tester/framework/cluster.py
geokala: OK, if we depend on the ID in other places then I'll modify it here.

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
geokala: Hmm... skip_community vs community_skip... no reason. You're right though, skip_<reason> is probably better, I'll tweak it.

I think I should be able to skip the entire module too, yes- I'll try it.

cosmo_tester/resources/attributes.yaml
geokala: Not env vars I think- we went there before and it was painful.

Anyway, if we want to test a community image then the config needs modifying anyway (to set the image name). I think that's more under the remit of the config changes either way?

cosmo_tester/framework/cluster.py
geokala: Ah- I wonder if this is the cause of the community bootstrap test problems- I'll have a poke.

cosmo_tester/framework/cluster.py
geokala: I'll put a comment in, but essentially it's because the built-in resources package (the default) should be a community one, unless we're making our manager blueprints deliberately unusable.

cosmo_tester/test_suites/snapshots/multi_tenant_snapshot_test.py
geokala: Agreed, and will probably be useful in the multi-tenant conversion stuff that I'll be stacking on after this.

PR #569, opened 2017-08-25
Cfy 7100 multi tenant tests

geokala:


cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
```diff
@@ -69,9 +66,10 @@ def test_inplace_upgrade(cfy,
     cfy.snapshots.upload([snapshot_path, '-s', snapshot_name])
     cfy.snapshots.restore([snapshot_name, '--restore-certificates'])
     _wait_for_restore(manager)
-    cfy.agents.install(['-a'])
```
geokala: This doesn't work with community (because it uses client.tenant.list()), and it doesn't work with premium, because the way the premium test is set up it creates deployments in two tenants, but none in default tenant. Jiras have been created for these issues.

cosmo_tester/test_suites/image_based_tests/hello_world_backwards_test.py
```diff
@@ -1,43 +0,0 @@
-########
```
geokala: These have been moved into the hello world tests and rationalised so that they're testing backwards compatibility of cloudify types versions rather than just extra arbitrary blueprints.

mcouthon:
Ok, even before starting, I have to say - a single commit with hundreds and hundreds of lines is no fun at all. Please consider splitting into multiple commits if possible (or at least in the future).

cosmo_tester/framework/examples/__init__.py
```diff
@@ -15,13 +15,13 @@
...
 import json
 import os
-import uuid
 from abc import ABCMeta
 
 import pytest
 import testtools
 
 from cosmo_tester.framework import git_helper
+from cosmo_tester.framework.util import set_client_tenant
```
mcouthon: Not sure if this make sense, but if you're doing a bunch of 
```
with set_client_tenant(self.manager, self.tenant)
```
Wouldn't it make sense to have this function be this class' method? Or maybe have a method that calls this function, just so you won't have to call `self.manager, self.tenant` each time.

cosmo_tester/framework/examples/__init__.py
```diff
@@ -134,15 +145,20 @@ def install(self):
     def clone_example(self):
         if not self._cloned_to:
             self._cloned_to = git_helper.clone(self.REPOSITORY_URL,
-                                               str(self.tmpdir),
+                                               os.path.join(
```
mcouthon: Maybe let's calculate this path in a separate line, and then used here. Seems wasteful to add 5 lines where 2 would suffice.

cosmo_tester/framework/util.py
```diff
@@ -485,3 +487,32 @@ def is_community():
...
+
+def get_test_tenant(test_param, manager, cfy):
+    if is_community():
+        tenant = 'default_tenant'
+        # It is expected that the plugin is already uploaded for the
+        # default tenant
+    else:
+        tenant = test_param
+        cfy.tenants.create(tenant)
+        manager.upload_plugin('openstack_centos_core',
```
mcouthon: It's a little odd that an action like uploading a plugin to the manager would be performed in an innocuously sounding func like `get_test_tenant `

cosmo_tester/resources/attributes.yaml
```diff
@@ -23,24 +23,26 @@ cloudify_manager_4_0_image_name: cloudify-manager-premium-4.0
 cloudify_manager_3_4_2_image_name: manager3.4.2-insecure-image
 notmanager_image_name: CentOS-7-x86_64-GenericCloud-1608
 
-centos7_username: centos
-centos7_image_name: CentOS-7-x86_64-GenericCloud-1608
+centos_7_username: centos
```
mcouthon: God bless you on all of these. I hated the lack of standardization

cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
```diff
@@ -43,17 +44,13 @@ def test_inplace_upgrade(cfy,
...
-    hello_world.blueprint_file = 'openstack-blueprint.yaml'
-    hello_world.inputs.update({
-        'agent_user': attributes.centos7_username,
-        'image': attributes.centos7_image_name,
-    })
-    hello_world.upload_blueprint()
-    hello_world.create_deployment()
-    hello_world.install()
-    hello_world.verify_installation()
+    hellos = get_hello_worlds(cfy, manager, attributes, ssh_key,
```
mcouthon: Why are we using `get_hello_worlds` here, and not the `hellos` fixture?

cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
mcouthon: Really? I was sure the part where no deps exist on the default tenant was taken care of. But apparently not. Can you link the Jiras here?

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
```diff
@@ -71,31 +72,40 @@ def cluster(
...
-
-    yield hw
-    if hw.cleanup_required:
-        logger.info('Hello world cleanup required..')
-        cluster.managers[0].use()
-        hw.cleanup()
-
-
-def test_data_replication(cfy, cluster, hello_world,
+def hello_worlds(cfy, cluster, attributes, ssh_key, tmpdir, logger):
```
mcouthon: I've already seen a `hello_worlds` fixture. Maybe we can differentiate this one by calling it something else?

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
```diff
@@ -71,31 +72,40 @@ def cluster(
...
+        hw.blueprint_file = 'openstack-blueprint.yaml'
+        hw.inputs.update({
+            'agent_user': attributes.centos_7_username,
+            'image': attributes.centos_7_image_name,
+        })
+        hws.append(hw)
+
+    yield hws
+    for hw in hws:
+        if hw.cleanup_required:
```
mcouthon: Why wasn't this `if` necessary in `get_hello_worlds`?

cosmo_tester/test_suites/image_based_tests/agent_installation_test.py
```diff
@@ -199,41 +82,69 @@ def test_ubuntu_trusty_provided_userdata_agent(cfy,
                                                tmpdir,
                                                logger):
     name = 'cloudify_agent'
-    user = attributes.ubuntu_username
+    os_name = 'ubuntu_14_04'
```
mcouthon: We don't have this in the attributes as well?

cosmo_tester/test_suites/image_based_tests/agent_installation_test.py
```diff
@@ -335,26 +317,30 @@ def _install_script(name, windows, user, manager, attributes, tmpdir, logger):
...
+        properties={'agent_config': {
+            'user': user,
+            'windows': windows,
+            'install_method': 'init_script',
+            'rest_host': manager.private_ip_address,
+            'broker_ip': manager.private_ip_address,
+            'name': name
+        }})
+    internal_ctx_dict = getattr(ctx, '_context')
+    internal_ctx_dict.update({
```
mcouthon: Why the move back to using this hack? What was wrong with providing the tenant and the token as arguments to `MockCloudifyContext`?

cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
geokala: The no deployments in default (current?) tenant is: https://cloudifysource.atlassian.net/browse/CFY-7214
The one that breaks it for community is addressed by: https://cloudifysource.atlassian.net/browse/CFY-7204

cosmo_tester/framework/examples/__init__.py
geokala: The problem is that it's used across a bunch of different tests and other components- e.g. it's used in the examples (which aren't parented off any common ancestor to the tests, thankfully), in the hostpool test, in the snapshots tests and image based tests.

The snapshot tests in particular don't use the class properties (and neither do some of the others), so I think trying to bake this into a common ancestor to everything would actually lead to some fairly unpleasant code.

cosmo_tester/framework/util.py
geokala: Yeah, I was slightly irked by that as well, but it also doesn't feel right as 'prepare_test_tenant', because if you're running community then it just returns one... but yes, I'll change it, probably to prepare_and_get_test_tenant. It's a bit more verbose but it does at least strongly imply that it's doing more than just retrieving some information.

cosmo_tester/resources/attributes.yaml
geokala: Yeah, it was quite irritating- and it also meant that there were a bunch of places we needed a pile of different methods rather than just one that substituted in OS names.

cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
geokala: I was asking myself the same question just now. It's because the inplace upgrade doesn't just have one manager (VM), it has two managers (well, VMs). Rather than hacking up a workaround to make the hello_worlds fixture work with both a single manager and a list of managers, I migrated the code out like this. I'll add a comment to the code to explain that.

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
geokala: This was just for the output- it shouldn't actually be required to make the tests work I think. Worth removing or not?

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
geokala: Yeah, I think maybe a little differentiation at least to make it clear which one is being viewed if someone has been wandering in a maze of twisty lines of code, all alike, for hours.

cosmo_tester/test_suites/image_based_tests/agent_installation_test.py
geokala: No, I changed it to have a username per OS (the way we did for centos and rhel even though they were the same username). This allowed me to DRY the code a lot more by having them all use the same set of inputs- e.g.
        'user': attributes['{os}_username'.format(os=os_name)],

cosmo_tester/test_suites/image_based_tests/agent_installation_test.py
geokala: Oh, that might work- that modification was one I was having to merge in and I must've missed what that other block of code was doing. I'll try it without this.

cosmo_tester/framework/examples/__init__.py
mcouthon: Yeah, fair enough

cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
mcouthon: Cool, sounds fair

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
mcouthon: Wait the cleanup isn't required, or the `cleanup_required` flag isn't?

cosmo_tester/test_suites/image_based_tests/agent_installation_test.py
mcouthon: 👍

cosmo_tester/test_suites/image_based_tests/agent_installation_test.py
mcouthon: Yeah, it should work. This was done a few days ago and the tests seem to pass with it

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
geokala: The check technically isn't required, since cleanup() checks whether that flag is set. In fact, it appears that the hello world actually logs this anyway... why on earth did I put that in there? I'll remove it.

PR #573, opened 2017-09-04
CFY-7154: Support overriding CLI URLs for tests

geokala:


cosmo_tester/framework/util.py
```diff
@@ -302,12 +302,48 @@ def get_plugin_wagon_urls():
...
+    if verification.status_code != 200:
+        raise RuntimeError(
+            error_base +
+            'Response to HEAD request was {status}'.format(
+                status=verification.status_code,
+            )
+        )
+
+
+def get_cli_package_url(platform):
```
geokala: I'm not actually sure why this was originally in the utils, but it does use a few other features from utils so I've left it there.

cosmo_tester/framework/util.py
```diff
@@ -302,12 +302,48 @@ def get_plugin_wagon_urls():
...
-    if is_community():
-        filename = 'cli-packages.yaml'
+def test_cli_package_url(url):
+    error_base = (
+        # Trailing space for better readability when cause of error
+        # is appended if there are problems.
+        '{url} does not appear to be a valid package URL. '
+    ).format(url=url)
+    try:
+        verification = requests.head(url)
```
tehasdf: maybe wanna `allow_redirects=True`? i'm not sure we want to also RuntimeError on a 304, instead of just following the redirect

cosmo_tester/framework/util.py
```diff
@@ -302,12 +302,48 @@ def get_plugin_wagon_urls():
...
+    config_cli_urls = get_attributes()['cli_urls_override']
+
+    if config_cli_urls.get(platform):
+        url = config_cli_urls[platform]
     else:
-        filename = 'cli-premium-packages.yaml'
-    return yaml.load(_get_package_url(filename))
+        if is_community():
+            filename = 'cli-packages.yaml'
+            packages_key = 'cli_package_urls'
```
tehasdf: used to be `packages`, not `package`

cosmo_tester/framework/util.py
geokala: It defaults to allow_redirects=True- I double checked that before I put it in. Is it worth specifying given that?

cosmo_tester/framework/util.py
geokala: Urgh. I'll change that and kick off a full community test. I thought I copy-pasted that.

cosmo_tester/framework/util.py
tehasdf: it doesn't according to the docs http://docs.python-requests.org/en/latest/user/quickstart/#redirection-and-history

geokala: That says it does default to True but that you can override it.

Also, from help(requests.request):
    :param allow_redirects: (optional) Boolean. Enable/disable GET/OPTIONS/POST/PUT/PATCH/DELETE/HEAD redirection. Defaults to ``True``.

And from running in an interpreter in the env with the tests:
>>> requests.get('https://www.google.com').status_code
200
>>> requests.get('https://www.google.com', allow_redirects=False).status_code
302

geokala:
Community tests are all running at the moment. Assuming no other code review issues I'll look to merge this tomorrow AM.

cosmo_tester/framework/util.py
tehasdf: Please note that you used `.get` to check, and `.head` in the actual code. From the docs:
>By default Requests will perform location redirection for all verbs except HEAD.

mcouthon: Yeah, Lukasz is right:
```
requests.head('https://www.google.com').status_code
302

requests.head('https://www.google.com', allow_redirects=True).status_code
200
```

PR #582, opened 2017-09-21
CFY-7159 rename cloudify cluster

funkyHat:


cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
```diff
@@ -18,16 +18,15 @@
...
 from . import get_hello_worlds
+from cosmo_tester.framework.test_hosts import TestHosts
 
 
 @pytest.fixture(scope='module')
 def cluster(request, cfy, ssh_key, module_tmpdir, attributes, logger):
     """Bootstraps a cloudify manager on a VM in rackspace OpenStack."""
     # need to keep the cluster to use its inputs in the second bootstrap
-    cluster = CloudifyCluster.create_bootstrap_based(
+    cluster = TestHosts.create_bootstrap_based(
```
mcouthon: Shouldn't that be changed to hosts as well, as is done elsewhere?

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
```diff
@@ -65,16 +65,16 @@ def cluster(
...
     finally:
-        cluster.destroy()
+        hosts.destroy()
 
 
 @pytest.fixture(scope='function')
 def ha_hello_worlds(cfy, cluster, attributes, ssh_key, tmpdir, logger):
     # Pick a manager to operate on, and trust the cluster to work with us
-    manager = cluster.managers[0]
+    manager = cluster.instances[0]
```
mcouthon: Ditto. If it's "hosts" everywhere, why is it "cluster" here?

funkyHat: Well, here it actually contains a cluster, so I left it... happy to change to `hosts` for consistency though

cosmo_tester/test_suites/ha/ha_negative_test.py
```diff
@@ -26,21 +26,21 @@
 
 
 @pytest.fixture(scope='function')
-def cluster(
+def hosts(
```
tehasdf: this could've been left named "cluster" since it actually is a cluster :)

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
```diff
@@ -29,28 +29,28 @@
 
 
 @pytest.fixture(scope='function', params=[2, 3])
-def cluster(
+def hosts(
```
tehasdf: same as the other - this actually is a cluster, so could've been just left as is

cosmo_tester/test_suites/ha/ha_negative_test.py
funkyHat: True :D I think consistency with the rest of the tests makes sense though, and also it means if someone searches for "cluster" in this file they'll find the references they probably actually care about

PR #583, opened 2017-09-25
CFY-7226 Make the multi network test verify upgrade as well

mcouthon:


cosmo_tester/framework/cluster.py
```diff
@@ -638,17 +641,17 @@ def _bootstrap_managers(self):
         super(BootstrapBasedCloudifyCluster, self)._bootstrap_managers()
 
         self._clone_manager_blueprints()
-        self._create_inputs_file()
-        self._bootstrap_manager()
+        for manager in self.managers:
+            self._create_inputs_file(manager)
+            self._bootstrap_manager()
```
tehasdf: this is very ugly because this depends on `_create_inputs_file` setting an instance attribute (inputs), and `_bootstrap_manager` reading that.
Perhaps make the inputs file a return from the first function, and an argument to the latter. Also of course would need to make the filename non-constant.
This would be a first step towards having parallel (threaded) bootstraps. (the next step would be having separate CFY_WORKDIRs)

cosmo_tester/test_suites/bootstrap_based_tests/multi_network_test.py
```diff
@@ -134,20 +128,34 @@ def multi_network_hello_worlds(cfy, managers, attributes, ssh_key, tmpdir,
...
+    {
+        "network_0": "10.0.0.6",
+        "network_1": "11.0.0.6",
+        "network_2": "12.0.0.6"
+    }
+    """
+
+    manager._logger.info('Adding extra NICs...')
+    for network_name, ip_addr in manager.networks.iteritems():
+        eth_num = network_name[-1]
```
tehasdf: this feels very fragile, is there a better way?

cosmo_tester/test_suites/bootstrap_based_tests/multi_network_test.py
```diff
@@ -134,20 +128,34 @@ def multi_network_hello_worlds(cfy, managers, attributes, ssh_key, tmpdir,
         hello.cleanup()
 
 
-def _enable_nics(manager, networks, tmpdir, logger):
-    """ Extra network interfaces need to be manually enabled on the manager """
+def _enable_nics(manager):
```
tehasdf: btw this should maybe probably be a method on the manager? it's one of the few (only?) real consumer of `.networks`, and all it does has to do with accessing attributes on the manager

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -91,11 +91,10 @@ def get_multi_tenant_versions_list():
...
 
 
-def upgrade_agents(cfy, manager, logger, tenants=('default_tenant',)):
+def upgrade_agents(cfy, manager, logger):
     manager.use()
-    for tenant in tenants:
-        _log('Upgrading agents', logger, tenant)
-        cfy.agents.install(['-t', tenant])
+    logger.info('Upgrading agents')
+    cfy.agents.install(['--all-tenants'])
```
tehasdf: is this actually ok? @geokala

cosmo_tester/test_suites/bootstrap_based_tests/multi_network_test.py
```diff
@@ -40,6 +40,7 @@
 IPADDR="{1}"
 NETMASK="255.255.255.128"
 DEFROUTE="no"
+MTU=1450
```
tehasdf: maybe leave a comment that this should be picked up by DHCP, but is hardcoded here.
Just in case in the future we run this in an env with MTU=1400, and this becomes a problem again

cosmo_tester/test_suites/snapshots/__init__.py
geokala: It wasn't on community tests- can you try running 'cfy agents install --all-tenants' on the latest community manager and see if it complains?

mcouthon: Trying now

PR #590, opened 2017-09-28
Replace CloudifyCluster.create_*_based with direct calls to subclasses

tehasdf:
The staticmethods buy us very little, but add a nontrivial amount
of complexity. They do nothing interesting other than just pass
through arguments, but provide no additional abstraction.

Instead of having a `create` argument for the image_based (and one
is going to also be needed for bootstrap_based), we can simply
stop using the staticmethods, but instead always use the right
class directly.

The only non-obvious thing in create_bootstrap_based was setting
the image name, which of course should live in the class itself,
because without it, the class won't work. (it doesn't make sense
to have a class that can only possibly work when created through
the specific factory)

### DIFF
--- a/cosmo_tester/framework/cluster.py
+++ b/cosmo_tester/framework/cluster.py
@@ -500,59 +500,6 @@ class CloudifyCluster(object):
     def _bootstrap_managers(self):
         pass
 
-    @staticmethod
-    def create_image_based(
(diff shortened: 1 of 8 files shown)


tehasdf:
Direct cause of this was, I'd also need a `create=False` in `create_bootstrap_based` for the MN+HA test, which is getting a bit ridiculous

PR #591, opened 2017-09-28
BootstrapBasedCluster: pass manager to _bootstrap_manager

tehasdf:
The function used to accept inputs, but it can just as well
accept a manager object and create the inputs.

This way it's easier for the caller, who now needs to only pass
the manager object, instead of creating the inputs file
(using the manager object) and passing that.

### DIFF
--- a/cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
+++ b/cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
@@ -65,7 +65,7 @@ def test_inplace_upgrade(cfy,
                    interval=1)
     cfy.snapshots.download([snapshot_name, '-o', snapshot_path])
     cfy.teardown(['-f', '--ignore-deployments'])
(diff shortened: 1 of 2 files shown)


cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
```diff
@@ -65,7 +65,7 @@ def test_inplace_upgrade(cfy,
                    interval=1)
     cfy.snapshots.download([snapshot_name, '-o', snapshot_path])
     cfy.teardown(['-f', '--ignore-deployments'])
-    cluster._bootstrap_manager()
```
tehasdf: this was wrong before, because `_bootstrap_manager` used to take an inputs file argument

funkyHat: Only since a few days ago ;)

geokala:

PR #593, opened 2017-10-03
CFY-7318 Update multi network test to add a network post-bootstrap

mcouthon:
This should verify that the manual mechanism of adding a network works
as expected

### DIFF
--- a/cosmo_tester/test_suites/bootstrap_based_tests/multi_network_test.py
+++ b/cosmo_tester/test_suites/bootstrap_based_tests/multi_network_test.py
@@ -176,9 +176,10 @@ def _add_new_network(manager, tmpdir, logger):
         ))
 
         logger.info('Recreating internal certs')
-        fabric_ssh.sudo('{python} {script} {ip}'.format(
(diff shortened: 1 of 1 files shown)


cosmo_tester/test_suites/bootstrap_based_tests/multi_network_test.py
```diff
@@ -101,11 +114,78 @@ def test_multiple_networks(managers,
...
+        logger.info('Updating the provider context...')
+        fabric_ssh.sudo('{python} {script} --networks {networks} {ip}'.format(
+            python=restservice_python,
+            script=update_ctx_script,
+            networks=remote_metadata_path,
+            ip=private_ip
+        ))
+
+        logger.info('Recreating internal certs')
+        fabric_ssh.sudo('{python} {script} {ip}'.format(
```
tehasdf: maybe let's also use explicit `--metadata filename` here instead of relying on the default, so that if a user reads this test as an example of how to do this, they understand that this file can be passed here

PR #595, opened 2017-10-09
CFY-7305 Test restored users are fixed

geokala:
Will cause master -> master snapshot tests to fail unless https://github.com/cloudify-cosmo/cloudify-manager/pull/987 is merged.

### DIFF
--- a/cosmo_tester/framework/util.py
+++ b/cosmo_tester/framework/util.py
@@ -489,16 +489,28 @@ class YamlPatcher(object):
 
 @retrying.retry(stop_max_attempt_number=10, wait_fixed=5000)
 def assert_snapshot_created(manager, snapshot_id, attributes):
-    url = 'http://{ip}/api/{version}/snapshots/{id}'.format(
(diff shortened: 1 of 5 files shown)


cosmo_tester/framework/util.py
```diff
@@ -489,16 +489,28 @@ def _raise_illegal(prop_path):
...
-    snapshot = AttributesDict(r.json())
-    assert snapshot.status == 'created', 'Snapshot not in created status'
+    snapshots = manager.client.snapshots.list()
+
+    existing_snapshots = {
+        snapshot['id']: snapshot['status']
+        for snapshot in snapshots
+    }
+
+    assert snapshot_id in existing_snapshots.keys(), (
```
tehasdf: no need for the `.keys()`, just `snapshot_id in existing_snapshots` is fine

geokala: Yeah, I prefer to be explicit <smiles>

PR #607, opened 2017-10-19
git_helper: do a shallow clone

tehasdf:
We don't need the whole history, so this should speed up the cloning
for free.


geokala:

PR #611, opened 2017-10-23
Small system tests improvements

mcouthon:


cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
```diff
@@ -54,64 +53,18 @@ def test_inplace_upgrade(cfy,
                               module_tmpdir, logger)
     for hello_world in hellos:
         hello_world.upload_and_verify_install()
-    create_snapshot(manager, snapshot_name, attributes, logger)
```
tehasdf: can we now remove the `create_snapshot` and friends (restore etc), if we can also replace other call sites?

cosmo_tester/framework/examples/hello_world.py
```diff
@@ -85,3 +85,16 @@ def assert_webserver_running(self, http_endpoint):
...
+
+
+def centos_hello_world(cfy, manager, attributes, ssh_key, tmpdir, logger,
+                       tenant='default_tenant', suffix=''):
+    hello = HelloWorldExample(
+        cfy, manager, attributes, ssh_key, logger, tmpdir,
+        tenant=tenant, suffix=suffix)
+    hello.blueprint_file = 'openstack-blueprint.yaml'
+    hello.inputs.update({
+        'agent_user': attributes.centos_7_username,
```
tehasdf: this will need to be coordinated with #616

mcouthon: Yeah, I'll have to merge one by one anyway, and adapt the changes as they come.

cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
mcouthon: I'd rather not - these calls are making extra checks and validations that are relevant for snapshot tests, and I think we better leave those in.

PR #616, opened 2017-11-15
CFY-7522 Use the install RPM and RHEL images in system tests

mcouthon:
* Replace manager/agent images to RHEL 7.4 (instead of Centos 7)
* Use the new install RPM when bootstrapping managers


mcouthon:
The RHEL part won't work for now - the RHEL images we have in RackSpace don't support `yum install` out of the box. IT are working on it.

cosmo_tester/framework/test_hosts.py
```diff
@@ -636,58 +635,62 @@ def _update_instances_list(self, outputs):
...
+                    self._manager_install_rpm,
+                    install_rpm_file
+                )
+            )
+            fabric_ssh.sudo('yum install -y {0}'.format(install_rpm_file))
+            fabric_ssh.put(
+                inputs_file,
+                '/opt/cloudify-manager-install/config.yaml'
+            )
+            fabric_ssh.run('cfy_manager install')
```
tehasdf: this doesn't need sudo, right?

cosmo_tester/test_suites/bootstrap_based_tests/inplace_upgrade_test.py
```diff
@@ -61,7 +61,8 @@ def test_inplace_upgrade(cfy,
 
     cfy.snapshots.download([snapshot_name, '-o', snapshot_path])
     cfy.teardown(['-f', '--ignore-deployments'])
-    hosts._bootstrap_manager(hosts._create_inputs_file(manager))
+    config_file = hosts._create_config_file(manager)
+    hosts._bootstrap_manager(manager, config_file)
```
tehasdf: i didn't believe it at first either, but yes, this is the only call site of `._bootstrap_manager`

cosmo_tester/framework/test_hosts.py
mcouthon: Yes. This depends on the user having sudo permissions (which we validate), but doesn't need to run as sudo.

PR #617, opened 2017-11-16
CFY-7552 Make stage work after restore

dmitryAII:


cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -678,21 +686,21 @@ def get_plugins_list(manager, tenant=None):
                 item['distribution'],
             )
             for item in manager.client.plugins.list()
-        ]
```
geokala: This isn't required for this commit is it- it won't fail flake8 and it's just causing noise in the PR.

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -647,6 +646,15 @@ def check_deployments(manager, old_deployments, logger,
     _log('Found correct deployments', logger, tenant)
 
 
+def check_status(manager):
+    status = manager.client.manager.get_status()
+    for service in status['services']:
+        for instance in service['instances']:
+            if instance['state'] != 'running':
+                raise Exception('Service %s is in status %s',
```
tehasdf: this needs to be formatted here, because `Exception` won't do it automatically
like for example:
```python
raise Exception('Service {0} is in status {1}'.format(instance, instance['state']))
```

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -647,6 +646,15 @@ def check_deployments(manager, old_deployments, logger,
     _log('Found correct deployments', logger, tenant)
 
 
+def check_status(manager):
+    status = manager.client.manager.get_status()
+    for service in status['services']:
+        for instance in service['instances']:
```
geokala: We already have code to check this on the managers- <host>.verify_services_are_running(), which handles community as well by not failing when stage is not running on community.

dmitryAII: It's not exactly what I need. This function waits for vm creation

cosmo_tester/test_suites/snapshots/__init__.py
dmitryAII: Fixed

cosmo_tester/test_suites/snapshots/__init__.py
dmitryAII: Fixed

cosmo_tester/test_suites/snapshots/__init__.py
geokala: It doesn't wait for VM creation, though it is also used for that.

If you want it without the retry it might be worth splitting the service status check there into a 'check_services' and then calling that, because otherwise you're still in the state where you're going to break community tests.

PR #621, opened 2017-11-27
SSL test: copy the external cert from the manager

tehasdf:
In the bootstrap process, we no longer do this automatically


geokala:

PR #625, opened 2017-12-05
CFY-7624 Fix some system tests

mcouthon:


cosmo_tester/resources/terraform/openstack-linux-cli-test.tf
```diff
@@ -81,6 +82,12 @@ resource "openstack_compute_secgroup_v2" "security_group" {
     ip_protocol = "tcp"
     cidr = "0.0.0.0/0"
   }
+  rule {
```
geokala: Can we add a comment to say what this rule is for?

cosmo_tester/resources/terraform/openstack-linux-cli-test.tf
```diff
@@ -159,10 +161,17 @@ resource "openstack_compute_instance_v2" "manager_server" {
...
+    destination = "/tmp/key.pem"
+  }
+
   provisioner "remote-exec" {
     inline = [
-      "echo hello world",
-      "sudo yum update openssl -y"
+      "echo Setting permissions for private key file: ${var.remote_key_path}",
+      "sudo cp /tmp/key.pem ${var.remote_key_path}",
+      "sudo chown cfyuser:cfyuser ${var.remote_key_path}",
```
geokala: <minor>
I tend to prefer:
sudo chown cfyuser: ${file_path}

This changes the user to the user specified and the group to that user's default group.

cosmo_tester/resources/terraform/scripts/linux-cli-test.sh
```diff
@@ -24,20 +20,18 @@ fi
...
-echo "public_ip: ${PUBLIC_IP}
-private_ip: ${PRIVATE_IP}
-ssh_user: ${MANAGER_USER}
-ssh_key_filename: ${PRIVATE_KEY_PATH}
-admin_username: admin
-admin_password: admin" > ${INPUTS_FILE_PATH}
-cat ${INPUTS_FILE_PATH}
+cfy profiles use ${PRIVATE_IP} -u admin -p admin -t default_tenant
+
+cfy blueprints upload https://github.com/cloudify-cosmo/cloudify-hello-world-example/archive/master.tar.gz -b bp -n singlehost-blueprint.yaml
```
geokala: I think it would be better for us to start using a test blueprint contained with the tests themselves, but we can do that separately (just flagging it here as a 'future-thought-relevant' thing).

cosmo_tester/resources/terraform/scripts/linux-cli-test.sh
```diff
@@ -24,20 +20,18 @@ fi
...
+cfy deployments create -b bp dep -i server_ip=${PRIVATE_IP} -i agent_user=${MANAGER_USER} -i agent_private_key_path=${AGENT_KEY_PATH}
+cfy executions start install -d dep
 
-echo "Setting permissions for private key file: ${PRIVATE_KEY_PATH}"
-chmod 400 ${PRIVATE_KEY_PATH}
+echo "Validating hello world is working..."
+# Because of set -e, this will fail if the string isn't found in the output
```
geokala: As we're depending on that, the set -e should probably be commented to indicate that it's critical to the tests.

cosmo_tester/resources/terraform/scripts/osx-cli-test.sh
```diff
@@ -18,20 +14,20 @@ echo $MACINCLOUD_PASSWORD | sudo -S installer -pkg /tmp/package.pkg -target /
...
 
-echo "Creating inputs file.."
-echo "public_ip: ${PUBLIC_IP}
-private_ip: ${PRIVATE_IP}
-ssh_user: ${MANAGER_USER}
-ssh_key_filename: ${PRIVATE_KEY_PATH}
-admin_username: admin
-admin_password: admin" > ${INPUTS_FILE_PATH}
-cat ${INPUTS_FILE_PATH}
+cfy=/usr/local/opt/cfy/bin/cfy
```
geokala: Given that this is the largely the same as the other code, is it possible for us to put this in another file and source it (or build the script from a jinja template perhaps?).

cosmo_tester/test_suites/bootstrap_based_tests/cli_test.py
```diff
@@ -33,117 +33,100 @@
...
 
     yield tester
 
     tester.perform_cleanup()
 
 
-def test_cli_on_centos_7(cli_package_tester, attributes):
-    cli_package_tester.inputs.update({
+@pytest.mark.parametrize('package_tester', ['linux'], indirect=True)
+def test_cli_on_centos_7(package_tester, attributes):
```
geokala: I wonder if we could parameterise this based on OS name and then provide a dict in the file which mapped out the inputs for each- that'd mean keeping one dict that contained just the extra inputs and then having one test runner that used pytest's parameterisation to pull the inputs from that dict for each OS in turn.

cosmo_tester/test_suites/image_based_tests/ssl_test.py
```diff
@@ -14,18 +14,22 @@
 #    * limitations under the License.
 
 from cosmo_tester.framework.examples.hello_world import centos_hello_world
-from cosmo_tester.framework.fixtures import bootstrap_based_manager
+from cosmo_tester.framework.fixtures import image_based_manager
```
geokala: While I'm not convinced our layout is as nice as it could be, while we have it in this structure we should probably make sure that something which isn't a bootstrap based test isn't in bootstrap_based_tests.

geokala: Ignore this, I misread the github header.

cosmo_tester/resources/terraform/scripts/linux-cli-test.sh
mcouthon: Agreed. We'll have a separate PR for that.
Maybe like what @funkyHat suggested we do instead of using hello world for sanity.

cosmo_tester/resources/terraform/scripts/osx-cli-test.sh
mcouthon: I considered it, but you need a separate script anyway, for platform specific stuff. You _could_ have a 3rd common script which both linux and OSX would source, but I also feel like it would warrant its own PR.

cosmo_tester/test_suites/bootstrap_based_tests/ssl_test.py
```diff
@@ -21,11 +21,15 @@
 
 manager = bootstrap_based_manager
 DEFAULT_TENANT_ROLE = 'user'
+REMOTE_EXTERNAL_CERT_PATH = '/etc/cloudify/ssl/cloudify_external_cert.pem'
 
 
 def test_ssl(cfy, manager, module_tmpdir, attributes, ssh_key, logger):
     cert_path = join(module_tmpdir, '.cloudify', 'profiles',
```
tehasdf: feels odd downloading it to the profile dir, maybe we should instead download it to just the tmpdir, since we're already calling `cfy profiles set -c` anyway?

cosmo_tester/test_suites/image_based_tests/ssl_test.py
```diff
@@ -14,12 +14,12 @@
...
 from cosmo_tester.framework.util import is_community
 from cloudify_rest_client.client import CloudifyClient
 from os.path import join
 
-manager = bootstrap_based_manager
+manager = image_based_manager
```
tehasdf: could do the import line `..import image_based_manager as manager` which would make this particular commit a net - in lines! ;P

cosmo_tester/resources/terraform/scripts/linux-cli-test.sh
tehasdf: `cfy blueprints upload cloudify-cosmo/cloudify-hello-world` is nicer than the whole github link

cosmo_tester/test_suites/bootstrap_based_tests/cli_test.py
```diff
@@ -38,112 +38,101 @@
...
+    package_tester.inputs.update({
         'cli_image': attributes.centos_7_image_name,
         'cli_user': attributes.centos_7_username,
         'manager_image': attributes.centos_7_image_name,
         'manager_user': attributes.centos_7_username,
         'cli_package_url': get_cli_package_url('rhel_centos_cli_package_url')
     })
-    cli_package_tester.run_test()
+    package_tester.run_test()
```
tehasdf: kinda sucks that we have the 2-step "update inputs + `.run_test()`". Looks like we should be able to make it `.run_test(inputs)`

cosmo_tester/resources/terraform/scripts/linux-cli-test.sh
mcouthon: Oh yeah, I forgot you can do that. Fill do

cosmo_tester/test_suites/image_based_tests/ssl_test.py
mcouthon: I like it a little better that way. For starters, you'd have to add a #NOQA otherwise, and I think it's a tiny bit more explicit that way

cosmo_tester/test_suites/image_based_tests/cli_test.py
```diff
@@ -312,5 +313,5 @@ def _copy_terraform_files(self):
             'terraform/scripts/osx-cli-test.sh'),
             self.tmpdir / 'scripts/osx-cli-test.sh')
 
-    def run_test(self):
-        super(_OSXCliPackageTester, self).run_test()
+    def run_test(self, inputs):
+        super(_OSXCliPackageTester, self).run_test(inputs)
```
tehasdf: this method can be removed, because all it does is run super (and hide the return value (which is none anyway))

PR #649, opened 2018-03-14
CFY-7877 Access InfluxDB directly via SSH using curl

mcouthon:
This is because now InfluxDB only binds on `localhost` (whereas it used
to bind on `0.0.0.0`), and we aren't able to access it via the manager's
public IP

### DIFF
--- a/cosmo_tester/framework/examples/nodecellar.py
+++ b/cosmo_tester/framework/examples/nodecellar.py
@@ -57,9 +57,6 @@ class NodeCellarExample(AbstractExample):
         self.assert_mongodb_collector_data()
 
     def assert_mongodb_collector_data(self):
-
-        influxdb = self.manager.influxdb_client
-
         # retrieve some instance id of the mongodb node
         mongo_node_name = 'mongod'
         with set_client_tenant(self.manager, self.tenant):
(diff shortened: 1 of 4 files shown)


cosmo_tester/framework/examples/nodecellar.py
```diff
@@ -68,18 +65,24 @@ def assert_mongodb_collector_data(self):
...
+                '"q=select sum(value) from '
+                '/{dep}\.{mongo_node}\.{instance_id}\.'
+                'mongo_connections_totalCreated/"'.format(
+                    url=self.manager.influxdb_url,
+                    dep=self.deployment_id,
+                    mongo_node=mongo_node_name,
+                    instance_id=instance_id
+                ), quiet=True
+            )
+            if result == '[]':
```
tehasdf: eh, maybe json.loads it after all? i suppose we want to know if the response format changes

cosmo_tester/framework/examples/nodecellar.py
```diff
@@ -68,18 +65,24 @@ def assert_mongodb_collector_data(self):
...
+                pytest.fail('Monitoring events for {0} node instance '
+                            'with id {1} were not found on influxDB'
+                            .format(mongo_node_name, instance_id))
```
tehasdf: and probably log the response

tehasdf:
related change that made influx localhost-only: cloudify-cosmo/cloudify-manager-install#146

cosmo_tester/framework/examples/nodecellar.py
mcouthon: The result is not necessarily in JSON format, I don't think

PR #663, opened 2018-04-19
CY-276 Default no metrics tests, except on heal

geokala:
Preparation for making the metrics+policy engine not be in the manager by default.


geokala:
My tests of this show everything passing except:
1. Deployment updates (looks to be due to recent change since it complains about --blueprint-id parameter)
2. All windows agent related tests are failing as the windows agent doesn't get installed. This is almost certainly due to a recent change in the agent code.
3. CLI test for OSX was not run as I don't have macincloud settings.

PR #679, opened 2018-05-08
Updated PyYAML version to fit openstacksdk reqs

qooban:
It should help to fix error during Stage-SystemTests job.


qooban:
Another PR created for that issue #680 .

PR #688, opened 2018-05-25
Use the new amqp client in the reboot tests

tehasdf:


tehasdf:
done the same in #689

mcouthon:
Except better, because I didn't create the client in a loop 😛

PR #748, opened 2018-08-29
CY-586 Add log saving for tests

davidginzbourg:
- Setting `CFY_LOGS_PATH_LOCAL` path as an env var will save logs of
test modules.
- Setting `SKIP_LOG_SAVE_ON_SUCCESS=True` will make sure not to save
logs when the tests are successful.

### DIFF
--- a/cosmo_tester/framework/test_hosts.py
+++ b/cosmo_tester/framework/test_hosts.py
@@ -138,6 +138,10 @@ class VM(object):
     def upload_necessary_files(self):
         return True
 
+    @property
+    def ssh_key(self):
+        return self._ssh_key
+
     image_name = ATTRIBUTES['default_linux_image_name']
     username = ATTRIBUTES['default_linux_username']
     branch_name = 'master'
@@ -635,9 +639,9 @@ class TestHosts(object):
         return self._attributes.manager_server_flavor_name
 
     def create(self):
-        """Creates the OpenStack infrastructure for a Cloudify manager.
+        """Creates the infrastructure for a Cloudify manager.
 
-        The openstack credentials file and private key file for SSHing
+        The credentials file and private key file for SSHing
         to provisioned VMs are uploaded to the server."""
         self._logger.info('Creating image based cloudify instances: '
                           '[number_of_instances=%d]', len(self.instances))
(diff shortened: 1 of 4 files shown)


cosmo_tester/framework/test_hosts.py
```diff
@@ -736,6 +746,38 @@ def _update_instances_list(self, outputs):
                     self._tmpdir
                     )
 
+    def _save_manager_logs(self, err_raised, purge=True):
+        self._logger.debug('_save_manager_logs started')
+        logs_dir = os.environ.get('CFY_LOGS_PATH_LOCAL')
+        test_path = self._tmpdir.name
+        if not logs_dir:
+            self._logger.debug('not saving manager logs')
```
tyacbovi: Explain why this did not happen, so the user could fix this with looking the code for the why

cosmo_tester/framework/test_hosts.py
```diff
@@ -736,6 +746,38 @@ def _update_instances_list(self, outputs):
...
+            return
+        if os.environ.get('SKIP_LOG_SAVE_ON_SUCCESS') and err_raised:
```
tyacbovi: I think it's enough to print this msg once when the flag is up, because you can not do anything with different log lines about not saving

cosmo_tester/framework/test_hosts.py
```diff
@@ -697,12 +701,18 @@ def create(self):
...
-    def destroy(self):
-        """Destroys the OpenStack infrastructure."""
-        self._logger.info('Destroying test hosts..')
-        with self._tmpdir:
-            self._terraform.destroy(
-                    ['-var-file', self._terraform_inputs_file, '-force'])
+    def destroy(self, err_raised=False):
+        """Destroys the OpenStack infrastructure.
+
+        :param err_raised: whether an error was raised.
```
tyacbovi: Add a little more info for the purpose of this flag, and lose the default because you will always have a value there

cosmo_tester/framework/fixtures.py
```diff
@@ -10,12 +10,14 @@ def image_based_manager(
...
             cfy, ssh_key, module_tmpdir, attributes, logger)
+    err_raised = True
     try:
         hosts.create()
         hosts.instances[0].use()
         yield hosts.instances[0]
+        err_raised = False
     finally:
-        hosts.destroy()
+        hosts.destroy(err_raised)
```
tyacbovi: Why not add a new function here that runs the "self._save_manager_logs(err_raised)"?
It is much cleaner because destroy doest suggest save logs.

davidginzbourg: @tyacbovi Because in other test/s I see that this code has been copied and therefore the common point is within the function [here for example](https://github.com/cloudify-cosmo/cloudify-system-tests/blob/CY-586/cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py#L66).
Having said that, I'll need to add this exception raised catch to all of them. Any other ideas?

cosmo_tester/framework/test_hosts.py
davidginzbourg: I removed the log output for when skipping saving on successful tests runs. I hope I understood what you meant.

cosmo_tester/framework/test_hosts.py
davidginzbourg: Added.

cosmo_tester/framework/test_hosts.py
davidginzbourg: Changed.

PR #750, opened 2018-09-06
Update openstack-vm.tf.template

opencm:


geokala:
Closed as this is presumably not required if we didn't do anything with it within six months.

PR #765, opened 2018-10-17
CY-698 New test to verify some common pitfalls of cfy_manager configure

mcouthon:


cosmo_tester/framework/test_hosts.py
```diff
@@ -140,6 +140,54 @@ def upload_necessary_files(self):
...
+
+    def get_remote_file_content(self, remote_path, use_sudo=True):
+        tmp_local_path = os.path.join(self._tmpdir, str(uuid.uuid4()))
+
+        self.get_remote_file(remote_path, tmp_local_path, use_sudo)
+
+        with open(tmp_local_path, 'r') as f:
+            content = f.read()
+
+        os.unlink(tmp_local_path)
```
geokala: These unlinks should probably be in a try|finally with an if os.path.exists guard?

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+from cosmo_tester.framework.fixtures import image_based_manager
+
+
+manager = image_based_manager
+
+REMOTE_CERT_PATH = '/etc/cloudify/ssl/cloudify_internal_ca_cert.pem'
+REMOTE_CONF_PATH = '/opt/manager/rest-security.conf'
+REMOTE_HOOKS_PATH = '/opt/mgmtworker/config/hooks.conf'
+LIST_USERS_COMMAND = 'rabbitmqctl -n cloudify-manager@localhost list_users'
```
geokala: Maybe LIST_MQ_USERS?

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -0,0 +1,85 @@
...
+    manager.put_remote_file_content(REMOTE_HOOKS_PATH, NEW_HOOKS)
+
+    logger.info('Running `cfy_manager configure`...')
+    manager.run_command('cfy_manager configure')
+
+    logger.info('Verifying certificates unchanged after configure...')
+    new_cert = manager.get_remote_file_content(REMOTE_CERT_PATH)
+    assert old_cert == new_cert
+
+    logger.info('Verifying RabbitMQ users recreated after configure...')
```
geokala: Should we also put a check in here that they're the same?
That'd require retrieving the (decoded) password after creating the extra tenant, but it would at least avoid us introducing bugs in future that effectively orphan all agents, so might be worth the effort now?

PR #783, opened 2018-12-02
Updating distributed installation system test with SSL certificates

barucoh:


cosmo_tester/framework/test_hosts.py
```diff
@@ -1000,11 +1014,89 @@ def database(self):
     def manager(self):
         return self.instances[1]
 
+    def _create_ssl_certificates_on_instances(self):
```
inbala: Pretty long function, can be split into smaller ones

cosmo_tester/framework/util.py
```diff
@@ -459,3 +465,220 @@ def mkdirs(folder_path):
...
+            subject_altips.append('IP:{name}'.format(name=name))
+
+    cert_metadata = ','.join([
+        ','.join(subject_altdns),
+        ','.join(subject_altips),
+    ])
+    return cert_metadata
+
+
+def _generate_ssl_certificate(ips,
```
inbala: Why is it private?

barucoh: Right, fixed

PR #786, opened 2018-12-06
Cy 634 add cloudify cluster support for external db implementations

barucoh:


cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
```diff
@@ -181,21 +181,27 @@ def test_failover(cfy, hosts, ha_hello_worlds, logger):
...
-def test_remove_manager_from_cluster(cfy, hosts, ha_hello_worlds, logger):
-    ha_helper.set_active(hosts.instances[1], cfy, logger)
-    ha_helper.delete_active_profile()
+def test_remove_manager_from_cluster(cfy, hosts, ha_hello_worlds, logger,
+                                     ignore_node_host_ip=None,
+                                     delete_profile=True):
+    ha_helper.set_active(hosts.instances[1], cfy, logger,
+                         ignore_node_host_ip=ignore_node_host_ip)
+    if delete_profile:
+        ha_helper.delete_active_profile()
```
tehasdf: delete_active_profile calls are likely not needed, retest without them and remove if passes

cosmo_tester/test_suites/bootstrap_based_tests/distributed_installation_test.py
```diff
@@ -112,38 +185,179 @@ def test_distributed_installation_scenario(database_and_manager,
...
-    logger.info('Creating secret agent_private_key_path as blueprint input')
-    cfy.secrets.create('agent_private_key_path', '-s',
-                       manager.remote_private_key_path, visibility='global')
 
-    logger.info('Creating secret host_ip as blueprint input')
-    cfy.secrets.create('host_ip', '-s', manager.ip_address,
-                       visibility='global')
+def reverse_cluster_test(cfy, cluster_machines, logger):
+    for manager in cluster_machines.instances:
+        disable_enable_node(manager, logger, disable=False)
```
tehasdf: i'd reverse the naming, making it `enable_disable_node(..., enable=True)` because the double negative in "disable=False" makes it harder to follow

also i guess maybe there can be a better name than "enable_disable" (maybe "toggle" or something)

cosmo_tester/test_suites/ha/ha_cluster_scenarios_test.py
barucoh: I had to put it there because the test deletes the active profile and then tries to use the joined nodes, which causes the command to fail because it is not the active node in the cluster.
I tried it with deleting the active profile and it failed.

cosmo_tester/test_suites/bootstrap_based_tests/distributed_installation_test.py
barucoh: Fixed

PR #789, opened 2018-12-12
CY-586 Add request param check before accessing param attribute

davidginzbourg:


cosmo_tester/framework/test_hosts.py
```diff
@@ -931,7 +931,8 @@ def _generate_prefix():
                     + <index_number_in_instances_list>__<terraform_id>__
             """
             prefix = '{}__'.format(instance.__class__.__name__)
-            if self._request:
+            if self._request \
+                    and getattr(self._request, 'param', None) is not None:
```
barucoh: Can be replaced with hasattr(self.request, 'param')

davidginzbourg: Replaced.

PR #791, opened 2018-12-13
CYBL-626 Add summary tests

geokala:


cosmo_tester/test_suites/summary/__init__.py
```diff
@@ -0,0 +1,599 @@
...
+    # Wait for any post restore commands to run
+    time.sleep(20)
+    yield manager
+
+
+class ExecutionWaiting(Exception):
+    """
+    raised by `wait_for_execution` if it should be retried
+    """
+    pass
```
tehasdf: you can omit the `pass` if it already has a docstring

cosmo_tester/test_suites/summary/__init__.py
```diff
@@ -0,0 +1,599 @@
...
+def retry_if_not_failed(exception):
+    return not isinstance(exception, ExecutionFailed)
+
+
+@retrying.retry(
+    stop_max_delay=5 * 60 * 1000,
+    wait_fixed=10000,
+    retry_on_exception=retry_if_not_failed,
+)
+def wait_for_execution(manager, execution, logger):
```
tehasdf: isn't there a `wait_for_execution` somewhere that we could use? if not, maybe this should go into a utils module

cosmo_tester/test_suites/summary/__init__.py
```diff
@@ -0,0 +1,599 @@
...
+
+    for summary_endpoint in (
+        prepared_manager.client.summary.blueprints,
+        prepared_manager.client.summary.deployments,
+        prepared_manager.client.summary.executions,
+        prepared_manager.client.summary.nodes,
+        prepared_manager.client.summary.node_instances,
+    ):
+        try:
+            summary_endpoint.get(_target_field='notafield')
```
tehasdf: perhaps this should use the CLI, so that part is also tested? system tests after all :P

cosmo_tester/test_suites/summary/__init__.py
```diff
@@ -0,0 +1,599 @@
...
+                        endpoint=summary_endpoint.summary_type,
+                        valid=', '.join(
+                            SUPPORTED_FIELDS[summary_endpoint.summary_type]
+                        ),
+                        message=str(err),
+                    )
+                )
+
+    if errors:
+        raise AssertionError('\n'.join(errors))
```
tehasdf: i guess can do `assert not errors`

cosmo_tester/test_suites/summary/__init__.py
```diff
@@ -0,0 +1,599 @@
...
+        '{0} != {1}'.format(results, expected)
+    )
+
+
+def test_basic_summary_deployments(prepared_manager):
+    results = sorted(prepared_manager.client.summary.deployments.get(
+        _target_field='blueprint_id',
+    ).items)
+
+    expected = sorted([
```
tehasdf: this feels wrong with no key func, how do dicts even sort?
(the answer is, they don't, and in py3 this sanely throws an exception instead)

geokala: Sorry, misunderstood. Either way, with the sorted this manages to be stable but without it doesn't. I could probably put in a key function if it's worthwhile?

cosmo_tester/test_suites/summary/__init__.py
geokala: Added.

cosmo_tester/test_suites/summary/__init__.py
geokala: Migrated to utils.

cosmo_tester/test_suites/summary/__init__.py
geokala: Done.

PR #811, opened 2019-02-06
CY-1019: system test

isaac-s:


cosmo_tester/resources/agent/winrm-agent-blueprint/plugins/mock-plugin/mock_plugin/tasks.py
```diff
@@ -14,15 +14,56 @@
...
+
+    ctx.instance.runtime_properties['userdata'] = userdata
 
 
 @operation
-def get_env_variable(env_variable, **_):
-    return os.environ[env_variable]
+def test_app(ctx, service_user, service_password, **_):
+    current_user = getpass.getuser()
+    if not service_user:
```
geokala: Not critical, but as a point of style I think it's generally better to do:
if <positive>:
  do thing
else:
  do other thing

Rather than starting with if not <thing>

isaac-s: You're right. Positive test first... I usually do this, not sure why I didn't here. I'll commit a change.

PR #821, opened 2019-03-06
Skip multi net cluster

geokala:


geokala:
Closing, as we'll remove this until we add the new one instead.

PR #834, opened 2019-04-21
4.6 merge master

adigrabow:


cosmo_tester/framework/test_hosts.py
```diff
@@ -254,6 +254,27 @@ def upload_necessary_files(self):
             fabric_ssh.sudo('chmod 440 {key_file}'.format(
                 key_file=REMOTE_PRIVATE_KEY_PATH,
             ))
+            self.enter_sanity_mode()
+
+    def enter_sanity_mode(self):
+        """
+        Test Managers should be in sanity mode to skip Cloudify license
+        validations.
+        :return:
```
tyacbovi: Can remove this

cosmo_tester/framework/test_hosts.py
```diff
@@ -432,7 +453,7 @@ def _create_config_file(self):
         config_file.write_text(install_config_str)
         return config_file
 
-    def bootstrap(self):
+    def bootstrap(self, db=False):
```
tyacbovi: What "db" means?

adigrabow: Whether or not this is an external DB machine (if it is we do not need to enter sanity mode)

cosmo_tester/framework/test_hosts.py
adigrabow: Why? I think it's a good comment :(

tyacbovi: I meant only line 263

cosmo_tester/framework/test_hosts.py
tyacbovi: so maybe change it to "with_db" or "db_contained"

PR #851, opened 2019-05-10
Agent install tests: add MANAGER_NAME envvar

tehasdf:


cosmo_tester/test_suites/image_based_tests/agent_installation_test.py
```diff
@@ -411,6 +411,9 @@ def _install_script(name, windows, user, manager, attributes, tmpdir, logger,
                                                defaults.INTERNAL_REST_PORT)
         ),
         constants.MANAGER_FILE_SERVER_ROOT_KEY: str(tmpdir),
+        constants.MANAGER_NAME: (
+            manager.client.manager.get_managers()[0].hostname
```
tehasdf: there's only one manager; but if there was more than one, at this point we should perhaps assume that it was replicated already anyway

PR #858, opened 2019-05-17
CY-1262 composer system tests fix

kubama:


kubama:
Good point, commit added

PR #873, opened 2019-06-13
Cy 1245 manager broker management

geokala:


cosmo_tester/test_suites/broker_management/conftest.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+@pytest.fixture()
+def broker(cfy, ssh_key, module_tmpdir, attributes, logger):
+    for _brokers in _get_brokers(cfy, ssh_key, module_tmpdir, attributes,
+                                 logger, instance_count=1):
+        yield _brokers[0]
+
+
+class JustHosts(BootstrapBasedCloudifyManagers):
```
tehasdf: i'd make a `MediumLinuxImage` in test_hosts, put the two methods (the ones other than bootstrap) from `BootstrapBasedCloudifyManagers` into it, have `BootstrapBasedCloudifyManagers` inherit that, and use that new class here directly

then we don't have the weird effect of "bootstrap based adds some behaviour, but here we also get rid of some of it"

and a more descriptive name i guess :P

cosmo_tester/test_suites/broker_management/conftest.py
```diff
@@ -0,0 +1,112 @@
...
+            'base': os.path.join(tempdir, node + '.'),
+        }
+    for paths in cert_paths.values():
+        paths['cert'] = paths['base'] + 'crt'
+        paths['key'] = paths['base'] + 'key'
+    util.generate_ca_cert(cert_paths['ca']['cert'], cert_paths['ca']['key'])
+
+    try:
+        for node in hosts.instances:
+            node.verify_services_are_running = skip
```
tehasdf: a better way would be to make the TestHosts subclass we use, use a VM class that doesnt implements all that shit, ie. probably the base `VM`. I don't currently see a good way to do that but maybe you will

cosmo_tester/test_suites/broker_management/conftest.py
```diff
@@ -0,0 +1,112 @@
...
+            node.upload_necessary_files = skip
+            node.upload_plugin = skip
+            node.enter_sanity_mode = skip
+
+        hosts.create()
+
+        for node_num, node in enumerate(hosts.instances, 1):
+            node_id = 'node' + str(node_num)
```
tehasdf: `node_id` is artificial and only used for `cert_paths`; i think i'd make `cert_paths` be a list, use the index directly here (or even `for node, certs in zip(hosts.instances, cert_paths)`, and make the CA pair just be local variables

simpler, no?

cosmo_tester/test_suites/broker_management/conftest.py
```diff
@@ -59,11 +67,13 @@ def _get_brokers(cfy, ssh_key, module_tmpdir, attributes, logger,
             node.verify_services_are_running = skip
             node.upload_necessary_files = skip
             node.upload_plugin = skip
-            node.enter_sanity_mode = skip
 
         hosts.create()
 
         for node_num, node in enumerate(hosts.instances, 1):
+            if node_num > broker_count:
```
tehasdf: this is also used below as `for node in hosts.instances[:broker_count]:` so i'd do, above, `brokers = hosts.instances[:broker_count]` and use it here and below

cosmo_tester/test_suites/broker_management/manager_test.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    brokers_list = list_brokers(manager)
+    assert len(brokers_list) == 1
+    assert expected_2 in brokers_list
+    logger.info('Removing broker succeeded.')
+
+
+def list_brokers(manager):
+    return json.loads(
+        manager.run_command(
```
tehasdf: i think we could do something with `with manager.ssh(pty=False, combine_stderr=False)` to avoid the 2 problems with the output

cosmo_tester/test_suites/broker_management/conftest.py
```diff
@@ -0,0 +1,112 @@
...
+                local_path=cert_paths[node_id]['key'],
+                remote_path='/tmp/rabbit.key',
+            )
+            node.put_remote_file(
+                local_path=cert_paths['ca']['cert'],
+                remote_path='/tmp/rabbit.ca',
+            )
+
+            node.additional_install_config = {
+                'rabbitmq': {
```
tehasdf: the two brokers aren't in a cluster?

cosmo_tester/test_suites/broker_management/manager_test.py
```diff
@@ -0,0 +1,79 @@
...
+    assert len(brokers_list) == 1
+    assert expected_1 in brokers_list
+    logger.info('Listing check passed.')
+
+    logger.info('Confirming add functionality.')
+    manager.run_command(
+        'cfy cluster brokers add {name} {ip} -n "{net}"'.format(
+            name=broker2.hostname,
+            ip=str(broker2.private_ip_address),
+            net='{{"default": {ip}, "testnet": "192.0.2.4"}}'.format(
```
tehasdf: i suppose at this point easier to use json.dumps here rather than format with the escaping

cosmo_tester/test_suites/broker_management/manager_test.py
```diff
@@ -0,0 +1,79 @@
...
+        'port': 5671,
+        'networks': {'default': str(broker2.private_ip_address),
+                     'testnet': '192.0.2.4'},
+        'name': broker2.hostname,
+    }
+
+    logger.info('Confirming list functionality.')
+    brokers_list = list_brokers(manager)
+    assert len(brokers_list) == 1
+    assert expected_1 in brokers_list
```
tehasdf: those two asserts are just `assert brokers_list == [expected_1]`. One line saved! (and same for the post-remove-check for two lines saved!)

cosmo_tester/test_suites/broker_management/manager_test.py
```diff
@@ -0,0 +1,79 @@
...
+        local_path=broker1.ca_path,
+        remote_path='/tmp/rabbit.ca',
+    )
+
+    manager.additional_install_config = {
+        'rabbitmq': {
+            'ca_path': '/tmp/rabbit.ca',
+            'cluster_members': {
+                broker1.hostname: {
+                    'default': str(broker1.private_ip_address),
```
tehasdf: it's weird that we'd need those str() everywhere around the private_ip_address, i don't think we need them

cosmo_tester/test_suites/broker_management/conftest.py
geokala: They don't start clustered- if they did, some of the tests would require removing and reinstalling.

cosmo_tester/test_suites/broker_management/manager_test.py
geokala: Apparently config.yaml gets upset when the IP address is !!python-unicode 192.0.2.56

cosmo_tester/test_suites/broker_management/manager_test.py
geokala: But the escape codes weren't in the stderr.

tehasdf: but hopefully they have to do with pty i'd imagine

PR #913, opened 2019-10-11
CY-1552 Test cluster with db cluster

geokala:


cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -104,34 +127,123 @@ def _get_hosts(cfy, ssh_key, module_tmpdir, attributes, logger,
             if pre_cluster_rabbit and node_num != 0:
                 join_target = brokers[0].hostname
 
+            if pre_cluster_rabbit:
+                print(dir(brokers[0]))
```
tehasdf: print

cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -154,21 +266,101 @@ def _get_hosts(cfy, ssh_key, module_tmpdir, attributes, logger,
...
+                if high_security:
+                    node.additional_install_config['postgresql_client'][
+                        'ssl_client_verification'] = True
+                    node.additional_install_config['postgresql_client'][
+                        'ssl_enabled'] = True
+                    node.additional_install_config['ssl_inputs'][
+                        'postgresql_client_cert_path'] = '/tmp/node.crt'
+                    node.additional_install_config['ssl_inputs'][
+                        'postgresql_client_key_path'] = '/tmp/node.key'
+            elif len(dbs) == 1:
```
tehasdf: this check could be more up front, before doing actual things like creating certs etc

cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -34,20 +34,25 @@ def brokers_and_manager(cfy, ssh_key, module_tmpdir, attributes, logger):
...
+                 logger):
     for _vms in _get_hosts(cfy, ssh_key, module_tmpdir,
                            attributes, logger,
-                           broker_count=3, db_count=3, manager_count=2):
+                           broker_count=3, db_count=3, manager_count=2,
+                           pre_cluster_rabbit=True):
         yield _vms
 
 
 def _get_hosts(cfy, ssh_key, module_tmpdir, attributes, logger,
```
tehasdf: maybe split this 500 line function? at least to making dbs and rabbits and managers in separate funcs?

cosmo_tester/test_suites/cluster/broker_management.py
```diff
@@ -19,6 +20,14 @@ def get_broker_listing(broker, prefix='rabbit@'):
     return brokers
 
 
+def manager_list_brokers(manager):
+    return json.loads(
+        manager.run_command(
+            'cfy cluster brokers list --json 2>/dev/null | cat'
```
tehasdf: why do we need cat, again?

geokala: Comment added to code- it's because of escape characters being added by cfy which I think we dug into when this was originally created.

cosmo_tester/test_suites/cluster/conftest.py
geokala: This is removed in one of the commits

cosmo_tester/test_suites/cluster/conftest.py
geokala: Refactored.

PR #917, opened 2019-10-31
CY-1875 Fix issue with using crt for mom system test

mabuaisha:


cosmo_tester/test_suites/manager_of_managers/regional_cluster.py
```diff
@@ -470,7 +473,7 @@ def _populate_status_output(self):
 
     def verify_installation(self):
         super(AbstractRegionalCluster, self).verify_installation()
-        assert self.master_ip
+        assert self.lb_ip
```
EarthmanT: i think endpoint or cluster_endpoint or endpoint_address or endpoint_ip would be better.

mabuaisha: Sure, will update it

cosmo_tester/test_suites/manager_of_managers/mom_test.py
```diff
@@ -267,29 +267,29 @@ def test_regional_cluster_with_floating_ip(
     first_cluster = floating_ip_2_regional_clusters[0]
     second_cluster = floating_ip_2_regional_clusters[1]
 
-    first_cluster.deploy_and_validate()
+    first_cluster.deploy_and_validate(timeout=6000)
```
tyacbovi: Could it take that long? If it did is it a good result?

cosmo_tester/test_suites/manager_of_managers/regional_cluster.py
```diff
@@ -15,12 +15,14 @@
 
 import os
 import json
+from contextlib import contextmanager
 
 from cloudify_cli.constants import DEFAULT_TENANT_NAME
+from fabric import api as fabric_api
```
tyacbovi: Please order by length and separate 3rd package from our own

cosmo_tester/test_suites/manager_of_managers/mom_test.py
mabuaisha: Actually this could around 4000 or more so I increased it up tp 6000, this will not affect the end result, but since we are doing a lot of operations like restore from snapshot and also create resources like tenants, plugins, blueprint, and even trigger workflow execution as part of the install workflow of the cluster,

cosmo_tester/test_suites/manager_of_managers/regional_cluster.py
mabuaisha: sure

PR #919, opened 2019-11-05
Cy 1537 add snapshot tests 5.0.5

glukhman:
DO NOT MERGE! not ready yet

### DIFF
--- a/cosmo_tester/framework/test_hosts.py
+++ b/cosmo_tester/framework/test_hosts.py
@@ -47,6 +47,7 @@ RSYNC_SCRIPT_URL = 'https://raw.githubusercontent.com/cloudify-cosmo/cloudify-de
 
 MANAGER_API_VERSIONS = {
     'master': 'v3.1',
+    '5.0.5': 'v3.1',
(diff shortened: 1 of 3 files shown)


cosmo_tester/framework/test_hosts.py
```diff
@@ -885,11 +886,20 @@ class Cloudify4_6Manager(_OldStatusFormat, _CloudifyManager):
...
+    branch_name = '5.0.5'
+
+
 class Cloudify4_6Manager_ClusterJoined(Cloudify4_6Manager):
     def verify_services_are_running(self):
         pass
 
 
+class Cloudify5_0_5Manager_ClusterJoined(Cloudify5_0_5Manager):
+    def verify_services_are_running(self):
```
tyacbovi: Why is this needed?

glukhman: Not sure. I figured out that it's needed for cluster to bypass the fact that the manager is not an AIO and doesn't have all services locally, but maybe I'm wrong about it. The framework is not very transparent

PR #921, opened 2019-11-11
CY-1910 Fix cluster system tests

inbala:
* The format of db and rabbit nodes in the config.yaml was changed.

* Update those changes in the cluster system tests.

### DIFF
--- a/cosmo_tester/test_suites/cluster/conftest.py
+++ b/cosmo_tester/test_suites/cluster/conftest.py
@@ -215,7 +215,9 @@ def _bootstrap_rabbit_node(node, rabbit_num, brokers, skip_bootstrap_list,
     if pre_cluster_rabbit:
         rabbit_nodes = {
             broker.hostname: {
-                'default': str(broker.private_ip_address),
+                'networks': {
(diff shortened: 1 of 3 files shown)


cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -348,18 +357,24 @@ def _bootstrap_manager_node(node, mgr_num, dbs, brokers, skip_bootstrap_list,
...
             'server_username': 'postgres',
             'server_password': dbs[0].pg_password,
         }
 
         if len(dbs) > 1:
-            node.additional_install_config['postgresql_server'][
-                'cluster'][
-                'nodes'] = [str(db.private_ip_address) for db in dbs
-                            if db.friendly_name not in skip_bootstrap_list]
+            db_nodes = {
```
geokala: Given that you're using this logic in a couple of places and it's now a step more complex than it used to be, would it be worth factoring it out?

geokala: Discussed in slack: Given the slight differences and this similar code only being in two places it's probably best left for the moment.

PR #946, opened 2020-01-23
Cy 2256 cluster queue failover test

glukhman:


cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -34,3 +39,148 @@ def test_cluster_single_db(cluster_with_single_db, logger, attributes, cfy):
...
+
+    # bring the stopped rabbit back up
+    with agent_broker.ssh() as broker_ssh:
+        broker_ssh.run('sudo service cloudify-rabbitmq start')
+
+    # verify that the resurrected rabbit shares the load
+    with new_agent_broker.ssh() as broker_ssh:
+        for broker in [broker1, broker2, broker3]:
+            num_queues = broker_ssh.run(
+                'sudo -u rabbitmq rabbitmqctl -n rabbit@{0} list_queues'
```
geokala: Doesn't this list all queues anyway? Is there a way to find out what the primary node for a queue is?

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -34,3 +39,148 @@ def test_cluster_single_db(cluster_with_single_db, logger, attributes, cfy):
...
+        verify_agent_broker_connection_and_get_broker_ip(mgr1)
+    assert new_agent_broker_ip != agent_broker_ip
+
+    new_agent_broker = None
+    for broker in [broker1, broker2, broker3]:
+        if broker.private_ip_address == new_agent_broker_ip:
+            new_agent_broker = broker
+            break
+
+    # spin up 3 deployments
```
geokala: Why are you spinning up three deployments here? The comment says what you're doing but not why.

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -34,3 +39,148 @@ def test_cluster_single_db(cluster_with_single_db, logger, attributes, cfy):
...
+    hello_world = centos_hello_world(cfy, mgr1, attributes, ssh_key,
+                                     logger, module_tmpdir)
+
+    _CloudifyManager.upload_necessary_files(mgr1)
+    _CloudifyManager.upload_necessary_files(mgr2)
+    _CloudifyManager.upload_plugin(mgr1,
+                                   mgr1._attributes.default_openstack_plugin)
+
+    # Install a hello world deployment.
+    # The test cluster has no outer network access so mgr1.use() won't do, so I
```
geokala: mgr1.use() might work if you make the install use the floating IPs instead of the private ones?

glukhman: I didn't want to change the install since it's hard to predict how it would affect all the other system tests. Though we might have to change it anyway once we add a LB to the framework (my next task)

cosmo_tester/test_suites/cluster/full_cluster_test.py
glukhman: I wanted to do some load-testing on the queues, similar to what Ahmad did. I'm not sure if it's good though so if you have better ideas...

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -34,3 +39,154 @@ def test_cluster_single_db(cluster_with_single_db, logger, attributes, cfy):
...
+        active_brokers = broker_ssh.run(
+            'sudo -u rabbitmq rabbitmqctl list_connections host'.format(
+                broker.hostname)).split('\r\n')[1:]
+        assert len(set(active_brokers)) == 2
+
+    # bring the stopped broker back up
+    with agent_broker.ssh() as broker_ssh:
+        broker_ssh.run('sudo service cloudify-rabbitmq start')
+
+    # spin up 3 deployments to check whether agents will pick up the new broker
```
geokala: If this entire section is intended to validate that agents can use all brokers then this is a fragile (the brokers are picked randomly so it's possible that not all of them will be picked) and expensive (lots of extra VMs deployed).

A better way might be to just use the deployment you already have, blocking access to all but one of the brokers, restarting the agent, then confirming it connected, and repeating that for each broker?

cosmo_tester/test_suites/cluster/full_cluster_test.py
geokala: Fair enough, I'll mark this as resolved.

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -34,3 +41,165 @@ def test_cluster_single_db(cluster_with_single_db, logger, attributes, cfy):
...
+    with agent_broker1.ssh() as broker_ssh:
+        broker_ssh.run('sudo service cloudify-rabbitmq stop')
+
+    # the agent should now pick another broker
+    validate_cluster_and_agents(mgr1, expected_broker_status='Degraded')
+    agent_broker_ip2 = \
+        verify_agent_broker_connection_and_get_broker_ip(mgr1)
+    assert agent_broker_ip2 != agent_broker_ip1
+
+    """
```
geokala: This should be comments (#), not a docstring because that only has relevance at the beginning of a function and is otherwise declaring and then throwing away a new string.

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -34,3 +41,165 @@ def test_cluster_single_db(cluster_with_single_db, logger, attributes, cfy):
...
+    for broker in [broker1, broker2, broker3]:
+        if broker.private_ip_address == agent_broker_ip2:
+            agent_broker2 = broker
+            break
+    with agent_broker2.ssh() as broker_ssh:
+        broker_ssh.run('sudo service cloudify-rabbitmq stop')
+    agent_broker_ip3 = \
+        verify_agent_broker_connection_and_get_broker_ip(mgr1)
+
+    # if the agent connected to the initial broker, we're done
```
geokala: This bit might be better written as: while not restart_broker_connected:

geokala: You could put a limit on it since it should take at most 2 more, so assert at the end of the loop that you haven't hit attempt 3 or something similar (to avoid the test just 'hanging' forever).

PR #955, opened 2020-02-06
Cy 2258 cluster manager failover test

glukhman:


cosmo_tester/test_suites/cluster/config/haproxy.cfg
```diff
@@ -0,0 +1,24 @@
...
+    timeout check 5s
+listen manager
+    bind *:80
+    bind *:443 ssl crt /etc/haproxy/cert.pem
+    redirect scheme https if !{ ssl_fc }
+    mode http
+    option forwardfor
+    stick-table type ip size 1m expire 1h
+    stick on src
+    option httpchk GET /api/v3.1/status HTTP/1.0\r\nAuthorization:\ Basic\ YWRtaW46YWRtaW4=\r\nTenant:\ default_tenant
```
geokala: Using admin:admin as the basic auth header will cause problems if we change the admin password in any cluster tests. We don't at the moment, but it's conceivably something that could happen. I'd suggest using the config from https://docs.cloudify.co/5.0.5/install_maintain/installation/installing-cluster/#management-service-load-balancer instead, which checks for a 401.

cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -60,11 +63,10 @@ def full_cluster(cfy, ssh_key, module_tmpdir, attributes,
 
 
 @pytest.fixture()
-def full_cluster_with_lb(cfy, ssh_key, module_tmpdir, attributes,
-                         logger):
+def cluster_with_lb(cfy, ssh_key, module_tmpdir, attributes, logger):
     for _vms in _get_hosts(cfy, ssh_key, module_tmpdir,
                            attributes, logger,
-                           broker_count=3, db_count=3, manager_count=2,
+                           broker_count=2, db_count=1, manager_count=3,
```
geokala: Depending on what is being tested with this cluster, you might be able to drop to one broker and two managers?

cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -489,36 +491,11 @@ def _bootstrap_lb_node(node, managers, tempdir, logger):
...
-        haproxy_config += '    server manager_{private} {public} maxconn ' \
-                          '100 ssl check-ssl port 443 ca-file /etc/haproxy/' \
-                          'ca.crt\n'.format(private=manager.private_ip_address,
-                                            public=manager.ip_address)
-    node.run_command('echo "{config}" | sudo tee '
-                     '/etc/haproxy/haproxy.cfg'.format(config=haproxy_config))
+    template = Environment(
+        loader=FileSystemLoader(CONFIG_DIR)).get_template('haproxy.cfg')
+    config = template.render(managers=managers)
+    node.run_command('echo \'{config}\' | sudo tee '
```
geokala: I think we have a method for uploading files somewhere, which would be better to use, in case we put quotes in the config.

PR #956, opened 2020-02-18
Cy 2306 test manager failover workflow resume

glukhman:


cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -183,6 +183,70 @@ def _wait_for_cfy_node_to_start_serving(cfy, timeout=15):
...
+                          '-b', plugin_name)
+    cfy.plugins.upload(path.join(blueprints_dir, plugin_wagon), '-y',
+                       path.join(blueprints_dir, plugin_yaml))
+    cfy.deployments.create(plugin_name, '-b', plugin_name)
+    _wait_for_deployment_creation(cfy, plugin_name)
+    execution_start_time = time.time()
+    mgr1.client.executions.start(plugin_name, 'install')
+
+    # check which management worker handles the execution
+    exec_node_id = json.loads(
```
tehasdf: at this point i'd just use the rest client instead of the cli, to avoid the `strip`s (there's `mgr1.client`)

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -183,6 +183,70 @@ def _wait_for_cfy_node_to_start_serving(cfy, timeout=15):
...
+    other_manager = None
+    for manager in [mgr1, mgr2]:
+        try:
+            manager.run_command(
+                'cat /var/log/cloudify/mgmtworker/mgmtworker.log '
+                '| grep {}'.format(exec_node_id))
+        except Exception:
+            continue
+        executing_manager = manager
+        other_manager = ({mgr1, mgr2} - {executing_manager}).pop()
```
tehasdf: cute :P

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -183,6 +183,70 @@ def _wait_for_cfy_node_to_start_serving(cfy, timeout=15):
...
+    exec_node_id = json.loads(
+        mgr1.run_command('cfy node-instances list --json').lstrip(
+            '\r\n\033[0m').rstrip('\r\n\033[0m]'))['id']
+    time.sleep(3)   # wait for mgmtworker to get the execution
+    executing_manager = None
+    other_manager = None
+    for manager in [mgr1, mgr2]:
+        try:
+            manager.run_command(
+                'cat /var/log/cloudify/mgmtworker/mgmtworker.log '
```
tehasdf: this can be written without cat, just `grep {} /var/log/blabla.log`

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -183,6 +183,70 @@ def _wait_for_cfy_node_to_start_serving(cfy, timeout=15):
     raise TimeoutException
 
 
+def test_workflow_resume_manager_failover(minimal_cluster, cfy):
+    broker, db, mgr1, mgr2 = minimal_cluster
+    mgr1.use(cert_path=mgr1.local_ca)
```
tehasdf: odd that we need to pass the cert path manually?

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -183,6 +183,70 @@ def _wait_for_cfy_node_to_start_serving(cfy, timeout=15):
...
+
+    # verify on the second manager that the execution completed successfully
+    executions = json.loads(cfy.executions.list('--json').stdout)
+    installs = [execution for execution in executions
+                if execution['workflow_id'] == 'install']
+    assert len(installs) == 1
+    assert installs[0]['status'] == 'completed'
+
+
+def _wait_for_deployment_creation(cfy, deployment_id, timeout=15):
```
tehasdf: unfortunate, we have something like this function in about 10000 places, we should make an official real one already

PR #959, opened 2020-03-06
Cy 2392 use blueprints for test infra

geokala:


cosmo_tester/framework/test_hosts.py
```diff
@@ -247,6 +243,8 @@ def create(
         os.makedirs(self._tmpdir)
         self._openstack = util.create_openstack_client()
         self.additional_install_config = {}
+        if upload_plugins is None:
```
tehasdf: is not?

cosmo_tester/framework/test_hosts.py
```diff
@@ -386,8 +384,14 @@ def server_id(self):
         key = 'server_id_{}'.format(self.index)
         return self._attributes[key]
 
-    @retrying.retry(stop_max_attempt_number=6 * 10, wait_fixed=10000)
     def verify_services_are_running(self):
+        if self.branch_name.startswith('4'):
```
tehasdf: eeww

cosmo_tester/framework/test_hosts.py
geokala: if self.upload_plugins is None:
(updated)

PR #960, opened 2020-03-13
Cy 2392 make test resources deploy with blueprints

geokala:


test-requirements.txt
```diff
@@ -3,3 +3,4 @@ git+https://github.com/cloudify-cosmo/cloudify-cli@master#egg=cloudify==master
 git+https://github.com/cloudify-cosmo/cloudify-agent@master#egg=cloudify-agent==master
 mock
 passlib==1.7.1
+git+https://github.com/cloudify-cosmo/cloudify-fabric-plugin@master#egg=cloudify-fabric-plugin==master
```
tehasdf: plz pin this and the cli to a specific tag

tehasdf:
this is too big to really review deeply, i mostly went over it for code-level correctness but some logic questions might still be there, but most of the questions are about what was there previously anyway

test-requirements.txt
geokala: CLI probably shouldn't be pinned, as it (at the moment) needs to be master I think?

geokala: Fabric pinned to 1.6.0.

PR #961, opened 2020-03-16
get_cli_version: use pkg_resources

tehasdf:
instead of a function from cli, which doesn't exist anymore

### DIFF
--- a/cosmo_tester/framework/util.py
+++ b/cosmo_tester/framework/util.py
@@ -32,6 +32,7 @@ from os import makedirs
 from tempfile import mkstemp
 from contextlib import contextmanager
 
+import pkg_resources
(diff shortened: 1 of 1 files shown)


cosmo_tester/framework/util.py
```diff
@@ -489,7 +490,7 @@ def is_redhat():
 def run(command, retries=0, stdin=b'', ignore_failures=False,
         globx=False, shell=False, env=None, stdout=None, logger=logging):
     def subprocess_preexec():
-        cfy_umask = 0022
+        cfy_umask = 0o22
```
tehasdf: unrelated but with the old `0022` syntax, my flake complains

PR #971, opened 2020-04-01
CY-2527 Upgrade fabric to 2.x

geokala:


cosmo_tester/framework/test_hosts.py
```diff
@@ -105,14 +105,18 @@ def linux_username(self, user):
...
+            host=self.ip_address,
+            user=self.linux_username,
+            connect_kwargs={
+                'key_filename': self.private_key_path,
+            },
+            port=22,
+            connect_timeout=3,
+        )
+        conn.open()
+        yield conn
```
mabuaisha, line 119: @geokala do not you think we should also handle closing the connection ?

```
    try:
        conn.open()
        yield conn
    except Exception as e:
         pass
    finally:
        conn.close()
```

cosmo_tester/framework/test_hosts.py
```diff
@@ -165,57 +169,73 @@ def upload_necessary_files(self):
...
+            'cp {} {}'.format(remote_path, remote_tmp),
+            use_sudo=True,
+        ),
 
         with self.ssh() as fabric_ssh:
             fabric_ssh.get(
-                remote_path,
+                remote_tmp,
                 local_path,
-                use_sudo=use_sudo
```
mabuaisha, line 175: @geokala just noticed that `use_sudo` is removed, is this because of the new fabric upgrade ?

geokala: It's removed because it was defaulted to True and never set to False anywhere, so seemed a bit pointless.

PR #976, opened 2020-04-05
Cy 2402 use pre installed installer rpm by tests

mabuaisha:


cosmo_tester/framework/test_hosts.py
```diff
@@ -814,9 +794,19 @@ def __init__(self,
...
         if bootstrappable:
             for instance in self.instances:
-                instance._image_name = ATTRIBUTES['default_linux_image_name']
+                instance._image_name = self.bootstrappable_image_name
                 instance.should_finalize = False
 
+    @property
+    def bootstrappable_image_name(self):
+        image_name = ATTRIBUTES['cloudify_manager_installer_image_name']
+        if not image_name:
```
geokala, line 803: I don't think this exception is needed?

cosmo_tester/framework/test_hosts.py
```diff
@@ -814,9 +794,19 @@ def __init__(self,
 
         if bootstrappable:
             for instance in self.instances:
-                instance._image_name = ATTRIBUTES['default_linux_image_name']
+                instance._image_name = self.bootstrappable_image_name
                 instance.should_finalize = False
 
+    @property
+    def bootstrappable_image_name(self):
+        image_name = ATTRIBUTES['cloudify_manager_installer_image_name']
```
geokala, line 802: Shouldn't this auto-generate the image name the way the manager image does (when the attribute isn't populated)?

Then you could just do a get on the attribute and not worry about exceptions on Jenkins.

cosmo_tester/framework/test_hosts.py
```diff
@@ -522,13 +522,8 @@ def _create_config_file(self, upload_license=True):
     def bootstrap(self, enter_sanity_mode=True, upload_license=False,
                   blocking=True, restservice_expected=True):
         self.restservice_expected = restservice_expected
-        manager_install_rpm = \
-            ATTRIBUTES.cloudify_manager_install_rpm_url.strip() or \
-            util.get_manager_install_rpm_url()
```
geokala, line 527: Do we use get_manager_install_rpm_url anywhere any more? If not, let's splat it.

cosmo_tester/framework/test_hosts.py
mabuaisha: Well, you are right for this and should be update once we sync the community version + premium version and thats the reason why I added the exception there and added a note about that

cosmo_tester/framework/test_hosts.py
mabuaisha: Manager of manager test use it and cannot drop it since need to download the rpm to the central manager image and drop the rpm to specific path so that spire plugin can use it

cosmo_tester/framework/test_hosts.py
mabuaisha: I replied to comment below

cosmo_tester/framework/test_hosts.py
mabuaisha: https://github.com/cloudify-cosmo/cloudify-system-tests/blob/master/cosmo_tester/test_suites/manager_of_managers/mom_test.py#L103

PR #986, opened 2020-04-13
Update hello world tests and fix requirements

geokala:


cosmo_tester/test_suites/image_based_tests/simple_deployment_test.py
```diff
@@ -0,0 +1,51 @@
...
+                                         image_based_manager,
+                                         cfy, upload=False)
+    image_based_manager.upload_test_plugin(tenant)
+    example = OnManagerExample(
+        cfy, image_based_manager, attributes, ssh_key, logger, tmpdir,
+        tenant=tenant
+    )
+
+    yield example
+    example.uninstall()
```
mabuaisha, line 51: Maybe it will be safer to add uninstall to finally with force removing 
```
try:
    yield example
finally:
    example.uninstall()

File: cosmo_tester/framework/examples/on_manager.py [hunk 2]
```diff
@@ -49,3 +53,22 @@ def check_files(self):
                 file_path = self.inputs['path'] + '_' + instance.id
                 content = self.manager.get_remote_file_content(file_path)
                 assert content == expected_content
+
+    def uninstall(self):
```
mabuaisha, line 57: Is it worth to add force flag and ignore error ?

cosmo_tester/test_suites/image_based_tests/simple_deployment_test.py
geokala: No need, these are only on-manager, so at the end the manager gets torn down anyway.

cosmo_tester/framework/examples/on_manager.py
geokala: No need- this is on-manager only. I'll look more into that for the off-manager one once I reach that point.

PR #991, opened 2020-04-16
CY-2393 Inplace upgrade test use new examples

geokala:


cosmo_tester/test_suites/image_based_tests/inplace_upgrade_test.py
```diff
@@ -31,47 +34,54 @@
...
+    hosts = Hosts(cfy, ssh_key, module_tmpdir, attributes, logger, 2)
+    hosts.instances[0] = get_image(request.param)
+    manager, vm = hosts.instances
+
+    manager.upload_files = False
+    manager.restservice_expected = True
+
+    vm.upload_files = False
+    vm._image_name = ATTRIBUTES['centos_7_image_name']
+    vm._linux_username = ATTRIBUTES['centos_7_username']
```
alexmolev, line 48: `centos_7_image_name` and `centos_7_username` can be moved to a constant

geokala: They're pulled from the configuration, but I am thinking about ways we might be able to make the whole TestHosts and VM stuff better.

PR #1016, opened 2020-05-26
Cy 2604 stop using cli

geokala:
This should be reviewed.


cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
     ])
 
 
-class ExecutionWaiting(Exception):
-    """
-    raised by `wait_for_execution` if it should be retried
-    """
+class ExecutionTimeout(Exception):
+    """Raised by `wait_for_execution` if the execution takes too long."""
```
tehasdf, line 405: Just `An execution timed out` would be good enough of a docstring: it's common for exception descriptions to be just the description of the problem.

Here's some examples from the standard library:
```python
class IOError(EnvironmentError)
 |  I/O operation failed.

class AssertionError(StandardError)
 |  Assertion failed.

class KeyboardInterrupt(BaseException)
 |  Program interrupted by user.
```

Objects are not generally expected to know anything about their users, so `ExecutionTimeout` cannot really know where is it going to be thrown.

Yes, I'm aware this is just a reformat, so it's fiiiine if you don't change it, but, I see a chance to make a line shorter! woo

cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
...
 
 
-@retrying.retry(
-    stop_max_delay=5 * 60 * 1000,
-    wait_fixed=10000,
-    retry_on_exception=retry_if_not_failed,
-)
-def wait_for_execution(manager, execution, logger):
+def wait_for_execution(manager, execution, logger, tenant=None,
+                       new_password=None, timeout=(5*60)):
```
tehasdf: the `timeout=(5*60)` is odd (parens?!) - maybe we can run CI's flake8 in a stricter mode? (this is E226, but that is ignored by flake8 by default

cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
...
+    # Timeout after ~5 minutes
+    timeout_time = current_time + timeout
+    password_updated = False
+    output_events(manager, execution, logger, to_time=current_time)
+
+    with set_client_tenant(manager, tenant):
+        while True:
+            try:
+                execution = manager.client.executions.get(execution['id'])
+            except UserUnauthorizedError:
```
tehasdf, line 433: can we please keep restore-specific bullshit out of generic functions like this one

cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
...
-        raise ExecutionFailed(
-            '{status}: {error}'.format(
-                status=execution.status,
-                error=execution['error'],
-            )
-        )
-    logger.info('Execution complete')
+    current_time = time.time()
+    # Timeout after ~5 minutes
+    timeout_time = current_time + timeout
```
tehasdf, line 425: use datetimes here:
```python
current_time = datetime.now()
timeout_time = current_time + timedelta(seconds=timeout)
```
and then instead of `convert_epoch_to_time_string`, you can just use `t.strftime(...)`, or perhaps even just `t.isoformat()`

cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
...
+    }
+    for event in events:
+        if event.get('type') == 'cloudify_event':
+            level = 'info'
+        else:
+            level = event.get('level')
+
+        if level not in log_methods:
+            logger.warn('Event level {} was unknown.'.format(level))
+            logger.warn('Event was: {}'.format(event))
```
tehasdf:
```suggestion
            logger.warn('Unknown event level: %s', level)
            logger.warn('Event was: %s', event)
```
the usual logger/format, but this also rewords the first message, because messages scan better if the thing you're looking for isn't between words (at the beginning would be even better, but at the end is also OK)

cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
...
+        else:
+            log_methods[level](event.get('message', 'Message not found'))
```
tehasdf: maybe something like `<MESSAGE NOT FOUND>`, so that it looks less like a real message

cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
...
+                    raise
+
+            prev_time = current_time
+            current_time = time.time()
+
+            output_events(manager, execution, logger, prev_time, current_time)
+
+            if time.time() >= timeout_time:
+                raise ExecutionTimeout(
+                    'Execution timed out in state: {status}'.format(
```
tehasdf, line 449: execution id? maybe workflow name?

cosmo_tester/test_suites/image_based_tests/test_tenant_creation_rabbitmq_down.py
```diff
@@ -1,29 +1,24 @@
...
-def test_tenant_creation_no_rabbitmq(manager):
-    with manager.ssh() as fabric:
-        fabric.sudo('systemctl stop cloudify-rabbitmq')
+def test_tenant_creation_no_rabbitmq(image_based_manager):
+    image_based_manager.run_command(
+        'systemctl stop cloudify-rabbitmq', use_sudo=True)
 
     try:
-        manager.client.tenants.create('badtenant')
+        image_based_manager.client.tenants.create('badtenant')
```
tehasdf, line 9: this is better written as `with pytest.raises(CloudifyClientError)`

cosmo_tester/framework/util.py
```diff
@@ -414,44 +402,113 @@ def generate_ca_cert(ca_cert_path, ca_key_path):
...
+def convert_epoch_to_time_string(inp):
+    if inp:
+        return time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(inp))
+    else:
+        return None
+
+
+def change_rest_client_password(manager, new_password):
+    manager.client = create_rest_client(manager.ip_address,
+                                        tenant='default_tenant',
```
tehasdf, line 534: tenant is already defaulted to this

cosmo_tester/test_suites/snapshots/__init__.py
```diff
@@ -110,14 +113,30 @@ def upload_snapshot(manager, local_path, snapshot_id, logger):
...
+    for snapshot in snapshots:
+        logger.info('%(id)s - %(status)s - %(error)s', **snapshot)
+
+
+def list_executions(manager, logger):
+    logger.info('Listing executions:')
+    executions = manager.client.executions.list(include_system_workflows=True)
+    for execution in executions:
+        logger.info('%(id)s (%(workflow_id)s) - %(status_display)s',
+                    **execution)
```
tehasdf: `**execution` is a TypeError here, has to be just `execution`

eg:
```python
>>> logging.info('asd %(x)s', x=5)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/logging/__init__.py", line 1629, in info
    root.info(msg, *args, **kwargs)
  File "/usr/lib/python2.7/logging/__init__.py", line 1167, in info
    self._log(INFO, msg, args, **kwargs)
TypeError: _log() got an unexpected keyword argument 'x'
>>> logging.info('asd %(x)s', {'x': 5})
INFO:root:asd 5
>>> 
```

tehasdf: ditto for other usages and also `**snapshot` etc

cosmo_tester/framework/util.py
geokala: Sure, let's make it prettier- pushed a commit for that.

cosmo_tester/framework/util.py
geokala: Updating to remove brackets. I think it was bracketed wherever I was refactoring it from, so I left it like that but yeah, not required here.

cosmo_tester/framework/util.py
geokala: Yup, migrating it out since it's only used in one place.

cosmo_tester/framework/util.py
geokala: I think I was considering timedeltas at the start of this, must've forgotten about it. Updated.

cosmo_tester/framework/util.py
geokala: %s already updated in a later commit. I'll update the log structure though, yes, good point.

cosmo_tester/framework/util.py
geokala: Yup, that's a better approach, updating.

cosmo_tester/framework/util.py
geokala: It uses id in a previous message. I'll add it there too.

cosmo_tester/test_suites/image_based_tests/test_tenant_creation_rabbitmq_down.py
geokala: So it is- will clean it up while we're here.

cosmo_tester/framework/util.py
geokala: It is- removing that line.

cosmo_tester/test_suites/snapshots/__init__.py
geokala: Huh, wonder how I missed that. Updated.

cosmo_tester/framework/test_hosts.py
```diff
@@ -588,7 +561,14 @@ def _new_wait_for_manager(self):
...
         if manager_status['status'] != HEALTHY_STATE:
             raise StandardError(
-                'Timed out: Reboot did not complete successfully'
+                'Timed out: Manager services did not start successfully. '
+                'Inactive services: {}'.format(
+                    ', '.join([
+                        item['extra_info']['systemd']['unit_id']
+                        for item in manager_status['services'].values()
+                        if item['status'] != 'Active'
+                    ])
```
tehasdf, line 577:
```suggestion
                    ', '.join(
                        item['extra_info']['systemd']['unit_id']
                        for item in manager_status['services'].values()
                        if item['status'] != 'Active'
                    )
```
(no `[]`)

cosmo_tester/framework/test_hosts.py
```diff
@@ -600,7 +580,10 @@ def _old_wait_for_manager(self):
                        service in ['postgresql', 'rabbitmq']):
                     if instance['state'] != 'running':
                         raise StandardError(
```
tehasdf, line 589: could well be a timeouterror :P

cosmo_tester/framework/test_hosts.py
geokala: Changing.

cosmo_tester/framework/test_hosts.py
geokala: Not until the test framework definitely works on py3, haven't got to that yet.

PR #1036, opened 2020-06-23
Cy 2730 add test for external resource

glukhman:


cosmo_tester/resources/blueprints/service_composition/shared_resource.yaml
```diff
@@ -18,18 +18,27 @@ inputs:
...
+      key_filename: { get_capability: [ { get_attribute: [ shared_resource_node, deployment, id ] }, agent_key ]}
 
 
 node_templates:
   shared_resource_node:
     type: cloudify.nodes.SharedResource
     properties:
+      client:
+        host: { get_input: client_ip }
+        username: admin
```
geokala, line 39: Should we make the username+password be inputs too and actually test that they're properly used?

cosmo_tester/test_suites/service_composition/shared_resource_test.py
```diff
@@ -20,3 +24,59 @@ def test_shared_resource(image_based_manager, ssh_key, logger, test_config):
         _verify_deployments_and_nodes(app, 2)
         _check_custom_execute_operation(app, logger)
         _verify_custom_execution_cancel_and_resume(app, logger)
+
+
+@pytest.fixture(scope='function')
+def managers(request, ssh_key, module_tmpdir, test_config, logger):
+    hosts = Hosts(ssh_key, module_tmpdir, test_config, logger, request, 2)
```
geokala, line 31: I think you need the 'passed' logic here (like the ones in the multi-tenant test), to make the conditional teardown work properly.

cosmo_tester/test_suites/service_composition/shared_resource_test.py
```diff
@@ -20,3 +24,59 @@ def test_shared_resource(image_based_manager, ssh_key, logger, test_config):
...
+        hosts.create()
+        yield hosts.instances
+    finally:
+        hosts.destroy()
+
+
+def test_external_shared_resource_idd(managers, ssh_key, logger, test_config):
+    tenant = 'test_external_shared_resource'
+    local_mgr, external_mgr = managers
+
```
geokala, line 42: So if you were to do something with the password, maybe changing the password (there's a util for it in snapshot tests currently I think), then adding a new account and using the new account would be good? That'd ensure it used both the username and password in case we did something really silly in the code in future.

geokala: (or made a naive mistake like making it always use the current credentials or something)

cosmo_tester/test_suites/service_composition/shared_resource_test.py
```diff
@@ -20,3 +24,59 @@ def test_shared_resource(image_based_manager, ssh_key, logger, test_config):
...
+    logger.info('Deploying the shared resource on an external manager.')
+    infra.upload_blueprint()
+    infra.create_deployment()
+    logger.info('Deploying an application which uses the resource, '
+                'on the local manager.')
+    app.upload_and_verify_install()
+
+    logger.info('Verifying inter-deployment dependency on both managers')
+    with util.set_client_tenant(app.manager.client, tenant):
+        local_idd = app.manager.client.inter_deployment_dependencies.list()[0]
```
geokala, line 57: Can't help but read this as local_if, probably have been to Cardiff (Caer Dydd) too often.

cosmo_tester/test_suites/service_composition/shared_resource_test.py
```diff
@@ -20,3 +24,70 @@ def test_shared_resource(image_based_manager, ssh_key, logger, test_config):
...
+        hosts.destroy(passed=passed)
+
+
+def test_external_shared_resource_idd(managers, ssh_key, logger, test_config):
+    tenant = 'test_external_shared_resource_idd'
+    local_mgr, external_mgr = managers
+
+    ext_username = 'dave'
+    ext_password = 'swordfish'
+    external_mgr.run_command('cfy users create -r sys_admin -p {0} {1}'.format(
```
geokala, line 49: This should probably be external_mgr.client.users.create(ext_username, ext_password, 'sys_admin') - best to avoid using the CLI where we don't need to.

PR #1038, opened 2020-06-30
CY-2945 - Update composer_test.py after changes in cloudify-composer package

qooban:


qooban:
@geokala Thanks for approval. After removing `grunt` and `bower` from Composer that change was a must. But despite that it was necessary it's now also less prone to errors and the control is more on composer side (as we can easily change what is to be executed under beforebuild, build, zip, upload scripts).

PR #1053, opened 2020-09-09
CY-3339 Avoid false failures when stopping vms

geokala:


cosmo_tester/framework/test_hosts.py
```diff
@@ -193,7 +194,7 @@ def stop(self):
             try:
                 self.run_command('echo Still up...')
                 time.sleep(3)
-            except SSHException:
+            except (SSHException, timeout):
```
tehasdf, line 197: looking at this line i'm gonna ask "what timeout?!". Especially since like this, it looks like some kinda variable
i think it would be better to do `import socket` and then `except socket.timeout`

geokala: Makes sense, update incoming.

tehasdf (approved):

PR #1068, opened 2020-09-30
CY-3598 Fix the three nodes cluster test

YoniItzhak:


cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -509,6 +509,8 @@ def _bootstrap_manager_node(node, mgr_num, dbs, brokers, skip_bootstrap_list,
     _add_monitoring_config(node, manager=True)
 
     # We have to block on every manager
+    with node.ssh() as fabric_ssh:
```
geokala: This commit isn't needed any more I think?

geokala (approved):

PR #1084, opened 2020-11-03
RD-323 Teardown cluster system test

YoniItzhak:
Writing a three nodes cluster teardown system test. The system test flow is:
1. Creating a 3 nodes Cloudify cluster. 
2. Creating an example deployment on this cluster. 
3. Removing the example deployment. 
4. Removing Cloudify from the cluster nodes. 
5. Reinstalling Cloudify on the nodes with new credentials and certificates. 

Note: We're not changing the Manager username and password. If we want to do this we also need to change node1 client headers, which without a proper setter can get messy.


cosmo_tester/framework/test_hosts.py
```diff
@@ -341,7 +342,7 @@ def assign(
         self.node_instance_id = node_instance_id
         self.deployment_id = deployment_id
         self.server_id = server_id
-        self.install_config = {
+        self.basic_install_config = {
```
geokala, line 345: Does the base one need to be an instance variable? If it wasn't then I think the copy.deepcopy could probably be safely removed too?

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -378,6 +382,94 @@ def test_three_nodes_cluster_status(three_nodes_cluster, logger):
...
+
+    for node in nodes_list:
+        node.install_config = copy.deepcopy(node.basic_install_config)
+
+    logger.info('Installing Cloudify cluster again')
+    run_cluster_bootstrap(nodes_list, nodes_list, nodes_list,
+                          skip_bootstrap_list=[], pre_cluster_rabbit=True,
+                          high_security=True, use_hostnames=False,
+                          tempdir=module_tmpdir, test_config=test_config,
+                          logger=logger, revert_install_config=True,
```
geokala, line 418: Maybe I'm misreading something, but it looks like you revert the install config /after/ this second bootstrap, which I think won't do anything- and you're reverting it to the basic (single node) install config, which wouldn't work if you did use it.

cosmo_tester/framework/test_hosts.py
YoniItzhak: Right, I haven't explained in the description of the PR the purpose of reverting the `install_config`. 
When we install a 3 nodes cluster, the same node is used for rabbitmq, DB, and Manager. During the bootstrap of each service (rabbitmq, DB, Manager), the `node.install_config` is updated with the values of the service currently being installed, but also keeps the values of the previous one. This means that after the cluster bootstrap, the `install_config` of each node has **all** values from each config file. This is why we need to revert the `install_config` of each node after installing each service in the 3 nodes case. 

We could delete the `basic_install_config` and put it in a global variable, but I don't think it would make things easier, i.e. we would still need `copy.deepcopy`.

Note: We could also revert the `install_config` by default for each node in the [run_cluster_bootstrap function](https://github.com/cloudify-cosmo/cloudify-system-tests/blob/1995acb59585eb1ced34fb23c6c0cd12f01a0c70/cosmo_tester/test_suites/cluster/conftest.py#L248), but I thought it might cause issues.

cosmo_tester/test_suites/cluster/full_cluster_test.py
YoniItzhak: I might be misunderstanding something, but each node's `install_config` is reverted before the second installation. Then, during the cluster bootstrap, after each service installation on each node, the `install_config` is reverted based on the explanation above.

cosmo_tester/framework/test_hosts.py
geokala: Hmm. OK, let's go with this for the moment and revisit it later once we do some more three node tests and get a feel for what's warty and what's necessary.

cosmo_tester/test_suites/cluster/full_cluster_test.py
geokala: Right, makes sense with your description above- I'm happyish with this for the moment, though the test is a little hard to follow but I think a lot of that is because the process itself is quite complex so I'll green it.

geokala (approved):

PR #1087, opened 2020-11-16
Rd 347 migrate cfy log tests to systest

glukhman:


cosmo_tester/config_schemas/platform_openstack.yaml
```diff
@@ -4,7 +4,7 @@ plugin_package_name:
   default: cloudify-openstack-plugin
 plugin_version:
   description: The expected version of this plugin.
-  default: 3.2.10
+  default: 3.2.18
```
geokala, line 7: Is .18 the latest version?

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +63,85 @@ def _test_cfy_install(run, example, paths, logger):
...
+
+
+def _test_cfy_logs(run, cli_host, example, paths, logger):
+    _set_ssh_in_profile(run, example, paths)
+
+    linebreak = '\\r\\n' if 'C:\\' in paths['cfy'] else '\n'
+    logger.info('Testing `cfy logs download`')
+    logs_dump_filepath = str(run(
+        '{cfy} logs download '.format(cfy=paths['cfy'])
+    ).stdout).split('archive to:')[1].split(linebreak)[0]
```
geokala, line 83: Does this command not provide json output?

geokala: (because we don't want to be doing horrible stuff like splitting on linebreaks- which I think .splitlines would do if we absolutely couldn't avoid it)

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +63,85 @@ def _test_cfy_install(run, example, paths, logger):
...
+
+    # Not working on windows!
```
geokala, line 85: So let's make cfy logs download retrieve a zip file.

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +63,85 @@ def _test_cfy_install(run, example, paths, logger):
...
+    local_logs_dump_filepath = os.path.join(tempfile.mkdtemp(), 'logs.tar')
+    cli_host.get_remote_file(logs_dump_filepath, local_logs_dump_filepath)
+    with closing(tarfile.open(name=local_logs_dump_filepath)) as tar:
```
geokala, line 88: Also, whichever archive type we use (and it should be zip so that windows users can also access it), I don't think we should test like this- let's instead stop all the cloudify services, then run the logs download both on the CLI VM and on the manager and check we get the same files.
Then, on the manager, we can check that the archive contains all the expected files by just comparing it to the target log files.

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +63,85 @@ def _test_cfy_install(run, example, paths, logger):
...
+        files = [f.name for f in tar.getmembers()]
+        assert 'cloudify/journalctl.log' in files
+        assert 'cloudify/nginx/cloudify.access.log' in files
+
+    logger.info('Testing `cfy logs backup`')
+    run('{cfy} logs backup --verbose'.format(cfy=paths['cfy']))
+    output = example.manager.run_command('ls /var/log').stdout
+    assert 'cloudify-manager-logs_' in output
```
geokala, line 96: What else is cfy logs backup supposed to do, because this checks that it creates a file/directory, not that it puts anything there.

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +63,85 @@ def _test_cfy_install(run, example, paths, logger):
...
+
+    logger.info('Testing `cfy logs purge`')
+    run('{cfy} logs purge --force'.format(cfy=paths['cfy']))
+    example.manager.run_command(
+        'test ! -s /var/log/cloudify/manager/cfy_manager.log')
```
geokala, line 101: Do we not empty any other logs when we run this?

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +63,85 @@ def _test_cfy_install(run, example, paths, logger):
...
+
+
+def _test_tmux_session(run, run_manager, example, paths, logger):
+    # this will go away after installing tmux in build-iamges pipeline
+    run_manager('sudo yum install tmux -y')
```
geokala, line 106: Just fail on it before then.

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +63,85 @@ def _test_cfy_install(run, example, paths, logger):
...
+    output = example.manager.run_command('ls /var/log').stdout
+    assert 'cloudify-manager-logs_' in output
+
+    logger.info('Testing `cfy logs purge`')
+    run('{cfy} logs purge --force'.format(cfy=paths['cfy']))
+    example.manager.run_command(
+        'test ! -s /var/log/cloudify/manager/cfy_manager.log')
+
+
+def _test_tmux_session(run, run_manager, example, paths, logger):
```
geokala, line 104: Do we actually need to provide this session functionality via cfy ssh? It seems like something that's unnecessary so maybe we could just remove it entirely and save ourselves the test.

cosmo_tester/test_suites/cli/__init__.py
geokala: Alternatively, we could create a stub fake-tmux and use that instead to check that the commands we'd execute do what we expect.

cosmo_tester/test_suites/cli/linux_test.py
```diff
@@ -44,11 +49,159 @@ def test_cli_install_flow_linux(linux_cli_tester, logger):
...
+                       paths,
+                       logger)
+
+
+def test_cfy_logs_linux_cluster(linux_cluster_cli_tester, logger):
+    cli_host = linux_cluster_cli_tester['cli_host']
+    manager = linux_cluster_cli_tester['manager']
+    paths = linux_cluster_cli_tester['paths']
+
+    linebreak = '\\r\\n' if 'C:\\' in paths['cfy'] else '\n'
```
geokala, line 84: This is in the linux tests, so it definitely doesn't need to check whether it's linux. Either way, hopefully we can use json output- and if we do need this function in multiple places let's factor it out into one place.

cosmo_tester/config_schemas/platform_openstack.yaml
glukhman: adopting the regex approach

cosmo_tester/test_suites/cli/__init__.py
glukhman: we need zip installed on the manager for that. Currently it isn't

cosmo_tester/config_schemas/platform_openstack.yaml
geokala: Are you intending that to be in this PR or another?

cosmo_tester/test_suites/cli/__init__.py
geokala: So to answer that question: Yes, it has --json output... which outputs nothing. Fantastic.

cosmo_tester/test_suites/cli/__init__.py
geokala: OK, in case anyone consults this later: We've had a bunch of back and forth about the windows side of this, and the upshot is that it's going to require some more work to handle windows with this so we're ignoring it for now.

cosmo_tester/test_suites/cli/__init__.py
geokala: And to update on this: The archive will indeed be different every time, so we can't test the way I suggested.

cosmo_tester/test_suites/cli/__init__.py
geokala: I believe @glukhman is following up with the team to see if this functionality is actually used.

cosmo_tester/config_schemas/platform_openstack.yaml
glukhman: In the current one

cosmo_tester/test_suites/cli/__init__.py
glukhman: That's the actual code it runs:
`        conn.run(
            'for f in $(sudo find /var/log/cloudify -name "*" -type f); '
            'do sudo truncate -s 0 $f; '
            'done'
        )`

cosmo_tester/config_schemas/platform_openstack.yaml
```diff
@@ -3,8 +3,8 @@ plugin_package_name:
   description: Package name of plugin for this platform. This plugin must be installed on the infrastructure manager with visibility=global.
   default: cloudify-openstack-plugin
 plugin_version:
-  description: The expected version of this plugin.
-  default: 3.2.10
+  description: The expected version of this plugin (can be a regex).
+  default: 3\.\d\.\d
```
geokala, line 7: I think this should be 3\.\d+\.\d+?

glukhman: ^3\.\d+\.\d+$, escaping the dots, of course

cosmo_tester/test_suites/cli/__init__.py
geokala: Huh... we should correct that at some point, probably to: sudo find /var/log/cloudify -type f -exec sudo truncate -s0 {}\;

Still, not for this PR.

geokala: So once again this one should probably be stopping services before it does the purge, then checking everything under that dir- e.g. I think this'll work (not tested):
sudo find /var/log/cloudify -type f test ! -s {} || echo {} && exit 1 \;

That should check each file under /var/log/cloudify is size zero; if not, it should echo the filename then exit nonzero, which should make the run_command fail.

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +62,101 @@ def _test_cfy_install(run, example, paths, logger):
...
+
+    # stop manager services so the logs won't change during the test
+    example.manager.run_command('cfy_manager stop')
+
+    logs_dump_filepath = [v for v in json.loads(run(
+        '{cfy} logs download --json'.format(cfy=paths['cfy'])
+    ).stdout.strip())['archive paths']['manager'].values()][0]
+
+    log_hashes = [f.split('  ')[0] for f in example.manager.run_command(
+        'find /var/log/cloudify -type f -exec md5sum {} \; | sort',  # NOQA
```
geokala, line 86: Why the noqa here?

glukhman:
Flake8 doesn't like \;

cosmo_tester/test_suites/cli/__init__.py
```diff
@@ -57,6 +62,101 @@ def _test_cfy_install(run, example, paths, logger):
...
+    _set_ssh_in_profile(run, example, paths)
+
+    # stop manager services so the logs won't change during the test
+    example.manager.run_command('cfy_manager stop')
+
+    logs_dump_filepath = [v for v in json.loads(run(
+        '{cfy} logs download --json'.format(cfy=paths['cfy'])
+    ).stdout.strip())['archive paths']['manager'].values()][0]
+
+    log_hashes = [f.split('  ')[0] for f in example.manager.run_command(
```
geokala, line 85: I think you can just do f.split()[0] - if no separator is supplied then it splits on any whitespace and removes empty strings from the result.

cosmo_tester/test_suites/cli/linux_test.py
```diff
@@ -44,11 +50,180 @@ def test_cli_install_flow_linux(linux_cli_tester, logger):
 def get_linux_image_settings():
     return [
         ('centos_7', 'rhel_centos_cli_package_url', 'rpm'),
-        ('ubuntu_14_04', 'debian_cli_package_url', 'deb'),
+        # ('ubuntu_14_04', 'debian_cli_package_url', 'deb'),
```
geokala, line 53: Why is this commented?

cosmo_tester/test_suites/cli/linux_test.py
```diff
@@ -44,11 +50,180 @@ def test_cli_install_flow_linux(linux_cli_tester, logger):
...
+    output = nodes[0].run_command('ls /var/log').stdout
+    assert 'cloudify-manager-logs_' in output
+
+    logger.info('Testing `cfy logs purge`')
+    cli_host.run_command('{cfy} logs purge --force'.format(cfy=paths['cfy']))
+
+    # Verify that each file under /var/log/cloudify is size zero
+    nodes[0].run_command(
+        'for f in $(sudo find /var/log/cloudify -type f -name "*") ; '
+        'do if $(test -s $f) ; then echo $f && exit 1 ; fi ; done'
```
geokala, line 139: This isn't quite correct, you've inverted the condition.
However, this seems to work in my testing:
find /var/log/cloudify -type f -exec test -s {} \; -print -exec false {} +

cosmo_tester/test_suites/cli/linux_test.py
```diff
@@ -44,11 +50,180 @@ def test_cli_install_flow_linux(linux_cli_tester, logger):
...
+            '-c {rest_cert}'.format(
+                ip=three_nodes_cluster[0].ip_address,
+                ssh_user=three_nodes_cluster[0].username,
+                ssh_key=remote_ssh_key_path,
+                rest_cert=remote_ca_cert_path,
+            )
+        )
+
+        # monkey with the profile so cluster nodes have an external ip,
+        # because the CLI host and the managers don't share an internal network
```
geokala, line 195: They should all be on the same network because only one network is deployed in any tests except multi-network ones. What are you seeing if you don't do this?

PR #1099, opened 2020-12-20
RD-837 Fix cfy_manager system test

YoniItzhak:


cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -63,7 +64,14 @@ def test_cfy_manager_configure(image_based_manager, logger, tmpdir,
     image_based_manager.put_remote_file_content(REMOTE_HOOKS_PATH, NEW_HOOKS)
 
     logger.info('Running `cfy_manager configure`...')
-    image_based_manager.run_command('cfy_manager configure')
+    image_based_manager.run_command(
```
geokala, line 67: Why do we do this? Am I missing the place where we're changing the password in this test?

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -14,13 +14,15 @@
...
 GET_MQ_PASSWORDS_CODE = '''
+import os
 import json
 
 from cloudify.cryptography_utils import decrypt
 
 from manager_rest.storage import models
 from manager_rest.flask_utils import setup_flask_app
 
+os.environ['MANAGER_REST_CONFIG_PATH'] = '/opt/manager/cloudify-rest.conf'
```
geokala, line 25: This shouldn't be in the tests

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: Well, the admin_password is blank in the config.yaml of the manager. Even though there is this section at the end:
```yaml
security:
  admin_username: admin
  admin_password: admin
```

The problem is that it's not under the `manager` section, and I couldn't find out where it is generated (maybe as part of the image generation?).

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: Without this line, it fails, so I wonder what would be the issue?

geokala: That doesn't make any sense- there is no /opt/manager/cloudify-rest.conf on the test VM- and this doesn't get exported to the remote nodes.

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
geokala: Base install config is generated here: https://github.com/cloudify-cosmo/cloudify-system-tests/blob/master/cosmo_tester/framework/test_hosts.py#L345

That puts it in the right place - I wonder if some of the config rewriting silliness is mangling it?

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
geokala: Marking this conversation as resolved- I didn't notice that this was actually in the GET_PASSWORDS code variable. D'oh.

geokala (approved):

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
geokala: We fixed this in the builder, marking resolved.

geokala (dismissed):

geokala (commented):
This can be green once we put the cfy_manager --private-ip stuff back in- sorry, I shouldn't have said to remove that.

geokala (approved):

PR #1103, opened 2020-12-21
RD-838 Fix snapshots inplace upgrade system test

YoniItzhak:
The snapshot inplace upgrade test fails while trying to uninstall the example deployment. This happens, as far as I understand, because the uninstall task is sent to the rabbitmq before it is initialized.
Waiting for the rabbitmq to be alive, should fix it -> well, it didn't.
Giving the manager more time to initialize (by increasing the sleep time) worked.

### DIFF
--- a/cosmo_tester/framework/test_hosts.py
+++ b/cosmo_tester/framework/test_hosts.py
@@ -651,6 +651,11 @@ class _CloudifyManager(VM):
                             )
                         )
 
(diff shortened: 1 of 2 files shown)


geokala (approved):

geokala:
Just noting that we're going with the kludge of increasing the sleep time to get the release finished (knowing that if someone is too quick to use an AIO on reboot they'll have to kill and re-do an operation), but this is something we'll investigate more post-release.

geokala (approved):

PR #1104, opened 2020-12-21
CR-834 use fixture for cluster test, cleanup archives between nodes

glukhman:


cosmo_tester/test_suites/cli/linux_test.py
```diff
@@ -102,6 +122,7 @@ def test_cfy_logs_linux_cluster(linux_cluster_cli_tester, tmpdir, logger):
                 [hashlib.md5(open(f.strpath, 'rb').read()).hexdigest() for f
                  in files if 'journalctl' not in f.basename])
             assert set(log_hashes) == set(log_hashes_local)
+            shutil.rmtree(str(tmpdir / 'cloudify'))
```
geokala, line 125: No need to delete it- pytest will deal with that.

glukhman: we do, this is between node logs in the same test. If the folder is dirty the assert will fail

geokala (approved):

PR #1110, opened 2020-12-29
RD-874 Add cfy_manager upgrade system tests

YoniItzhak:
This PR covers the cfy_manager upgrade on an AIO manager.

### DIFF
--- a/cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
+++ b/cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
@@ -1,0 +1,7 @@
+namespace: upgrade
+manager_image:
+  description: Name of image with 5.1.0 manager installer.
+  default: cfy-installer-premium-5.1.0-rhel
+upgrade_rpm_path:
+  description: The manager-install RPM path to upgrade to.
+  default: 'http://cloudify-release-eu.s3.amazonaws.com/cloudify/5.1.1/.dev1-release/cloudify-manager-install-5.1.1-.dev1.el7.x86_64.rpm'
--- a/cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
+++ b/cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
@@ -1,7 +1,4 @@
 namespace: upgrade
-manager_image:
-  description: Name of image with 5.1.0 manager installer.
-  default: cfy-installer-premium-5.1.0-rhel
 upgrade_rpm_path:
   description: The manager-install RPM path to upgrade to.
   default: 'http://cloudify-release-eu.s3.amazonaws.com/cloudify/5.1.1/.dev1-release/cloudify-manager-install-5.1.1-.dev1.el7.x86_64.rpm'
--- a/cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
+++ b/cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
@@ -1,7 +1,7 @@
 import json
 import pytest
 
-from cosmo_tester.framework.test_hosts import Hosts
+from cosmo_tester.framework.test_hosts import Hosts, get_image
 from cosmo_tester.framework.examples import get_example_deployment
 from cosmo_tester.framework.util import validate_cluster_status_and_agents
 
@@ -40,13 +40,10 @@ with open('%s', 'w') as f:
 
 @pytest.fixture(scope='function')
 def manager_5_1_0(request, ssh_key, module_tmpdir, test_config, logger):
-    hosts = Hosts(ssh_key, module_tmpdir, test_config, logger, request,
-                  bootstrappable=True)
-    hosts.instances[0].image_name = test_config['manager_image_names_rhel'][
-        '5_1_0_installer']
+    hosts = Hosts(ssh_key, module_tmpdir, test_config, logger, request)
+    hosts.instances[0] = get_image('5.1.0', test_config)
 
     hosts.create()
-    hosts.instances[0].bootstrap()
     try:
         yield hosts.instances[0]
     finally:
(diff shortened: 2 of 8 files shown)


cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
```diff
@@ -0,0 +1,7 @@
+namespace: upgrade
+manager_image:
```
geokala, line 2: This should not be here, we already have a list of images elsewhere- please use them.

geokala: Note that you can see the usage of this between the snapshot tests' __init__.py and conftest.py

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -33,6 +40,21 @@
...
 
 
+@pytest.fixture(scope='function')
+def manager_5_1_0(request, ssh_key, module_tmpdir, test_config, logger):
+    hosts = Hosts(ssh_key, module_tmpdir, test_config, logger, request,
+                  bootstrappable=True)
+    hosts.instances[0].image_name = test_config['upgrade']['manager_image']
+
+    hosts.create()
+    hosts.instances[0].wait_for_ssh()
```
geokala, line 50: You don't need to wait_for_ssh before calling bootstrap- it waits already.

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -33,6 +40,21 @@
...
+    hosts.instances[0].bootstrap()
```
geokala, line 51: Also, if you use the actual 5.1.0 image then you don't need to bootstrap at all.

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -98,6 +120,31 @@ def test_cfy_manager_configure(image_based_manager, logger, test_config):
...
+    validate_agents(manager_5_1_0, example.tenant)
+
+    logger.info('Installing new RPM')
+    manager_5_1_0.run_command('yum install -y {rpm}'.format(
+        rpm=test_config['upgrade']['upgrade_rpm_path']), use_sudo=True)
+    logger.info('Upgrading manager')
+    manager_5_1_0.run_command('cfy_manager upgrade -v')
+
+    _validate_manager_status(manager_5_1_0)
+    validate_agents(manager_5_1_0, example.tenant)
```
geokala, line 136: After this you should probably have it run example.uninstall() as this gives an even more clear indication that things are indeed fully functional.

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
```diff
@@ -98,6 +120,31 @@ def test_cfy_manager_configure(image_based_manager, logger, test_config):
...
+
+
+@retry(stop_max_attempt_number=30, wait_fixed=2000)
+def _validate_manager_status(manager):
```
geokala, line 140: It'd probably be better to factor '_wait_for_healthy_broker_cluster out from full_cluster_test into utils and use it here as well.

geokala: In fact, given that you're asserting on agent state as well, maybe it'd be best to factor out _validate_cluster_and_agents to utils as well and use it instead?

cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
YoniItzhak: We do have a list of images [here](https://github.com/cloudify-cosmo/cloudify-system-tests/blob/master/cosmo_tester/config_schemas/manager_image_names_rhel.yaml), but it doesn't include the 5.1.0-installer. 

Using the [5.1.0 image](https://github.com/cloudify-cosmo/cloudify-system-tests/blob/master/cosmo_tester/config_schemas/manager_image_names_rhel.yaml#L8) will fail because of the [issue described in RD-868](https://cloudifysource.atlassian.net/browse/RD-868). 

Should I instead add a 5.1.0-installer to `manager_image_names_rhel.yaml`?

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: Cool, I'll drop it. I was using [this fixture](https://github.com/cloudify-cosmo/cloudify-system-tests/blob/master/cosmo_tester/test_suites/bootstrap_based_tests/conftest.py#L16) for reference, so maybe it should be changed there as well?

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: Replied to it in my first reply.

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: Correct, I forgot about it. Thanks!

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: I absolutely think this is a better idea, and I actually thought about it too. The only thing is that I don't see any utils file that is used across test_suits, so I thought there is a reason for it. Should I add one?

cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
geokala: How about updating the config.yaml on the 5.1.0 image so that the broken entries are correct?

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
geokala: That fixture doesn't bootstrap, which is why it does wait_for_ssh

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
geokala: Hmm... there should be stuff in the framework folder I think? If not, yes, make one.

cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
YoniItzhak: I talked with @mabuaisha and this is not possible.

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: Makes sense.

cosmo_tester/test_suites/image_based_tests/cfy_manager_test.py
YoniItzhak: Cool, I will.

cosmo_tester/config_schemas/cfy_manager_upgrade.yaml
YoniItzhak: In any case, I'll add the 5.1.0-installer to `manager_image_names_rhel.yaml` and `manager_image_names_centos.yaml` for future use.

geokala (approved):

PR #1111, opened 2020-12-30
RD-874 Add cluster upgrade system tests

YoniItzhak:
This PR currently takes care only of the 3 nodes case. Once it is approved, I'll continue for the 9 nodes.

### DIFF
--- a/cosmo_tester/framework/util.py
+++ b/cosmo_tester/framework/util.py
@@ -663,17 +663,15 @@ def update_dictionary(dict1, dict2):
 
 def validate_cluster_status_and_agents(manager,
                                        tenant,
+                                       logger,
                                        expected_brokers_status='OK',
                                        expected_managers_status='OK',
-                                       agent_validation_manager=None,
-                                       logger=None):
+                                       agent_validation_manager=None):
     if not agent_validation_manager:
         agent_validation_manager = manager
-    if logger:
-        logger.info('Validating agents')
+    logger.info('Validating agents')
     validate_agents(agent_validation_manager, tenant)
-    if logger:
-        logger.info('Validating cluster status')
+    logger.info('Validating cluster status')
     validate_cluster_status(manager,
                             expected_brokers_status,
                             expected_managers_status)
--- a/cosmo_tester/framework/util.py
+++ b/cosmo_tester/framework/util.py
@@ -665,10 +665,15 @@ def validate_cluster_status_and_agents(manager,
                                        tenant,
                                        expected_brokers_status='OK',
                                        expected_managers_status='OK',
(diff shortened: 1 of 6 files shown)


cosmo_tester/framework/util.py
```diff
@@ -665,10 +665,15 @@ def validate_cluster_status_and_agents(manager,
                                        tenant,
                                        expected_brokers_status='OK',
                                        expected_managers_status='OK',
-                                       agent_validation_manager=None):
+                                       agent_validation_manager=None,
+                                       logger=None):
```
geokala, line 669: I think the logger should be mandatory- it makes it more clear where failures occur when you have preceding lines saying what was being attempted.

cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -177,6 +188,13 @@ def _get_hosts(ssh_key, module_tmpdir, test_config, logger, request,
         for node in hosts.instances:
             node.verify_services_are_running = skip
 
+        if installer_image_name:
+            # The cluster images distro is rhel by default, so let's keep it
```
geokala, line 192: It's centos if you specify to use centos in the test config, so we should support that.

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -402,6 +404,59 @@ def test_three_nodes_cluster_teardown(three_nodes_cluster, ssh_key,
     _assert_cluster_status(node1.client)
 
 
+def test_three_nodes_cluster_upgrade(three_nodes_5_1_0_cluster, ssh_key,
+                                     test_config, logger):
+    node1, node2, node3 = three_nodes_5_1_0_cluster
+    nodes_list = [node1, node2, node3]
+    logger.info('Installing example deployment')
+    example = get_example_deployment(node1, ssh_key, logger,
+                                     'cluster_teardown', test_config)
```
geokala, line 413: I'm not convinced 'cluster_teardown' is the right name for this?

YoniItzhak: You're absolutely right, I missed it.

cosmo_tester/test_suites/cluster/conftest.py
YoniItzhak: Thanks, will add a centos support.

cosmo_tester/framework/util.py
YoniItzhak: Agreed, I'll add `logger` to all places that use the function.

geokala (approved):

PR #1116, opened 2021-01-03
Update manager images for system tests

mabuaisha:


YoniItzhak (approved):

cosmo_tester/config_schemas/manager_image_names_rhel.yaml
```diff
@@ -1,11 +1,17 @@
...
+  default: cloudify-manager-premium-5.2.0.dev1-rhel
+"installer":
   description: Name of image with manager installer for bootstrap or cluster installs.
+  default: cfy-installer-premium-5.2.0.dev1-rhel
+"5_1_1_installer":
+  description: Name of image with 5.1.1 manager installer for bootstrap or cluster installs.
   default: cfy-installer-premium-5.1.1-rhel
-5_1_0_installer:
+"5_1_1":
+  description: Name of manager image for 5.1.1. All non master images are only used for upgrade (snapshot) tests.
```
YoniItzhak, line 12: It is true that they are used only for "upgrade". But now with `cfy_manager upgrade`, this is not only snapshots. So "(snapshots)" needs to be deleted.

cosmo_tester/config_schemas/manager_image_names_centos.yaml
```diff
@@ -1,11 +1,17 @@
...
+  default: cloudify-manager-premium-5.2.0.dev1
+"installer":
   description: Name of image with manager installer for bootstrap or cluster installs.
+  default: cfy-installer-premium-5.2.0.dev1
+"5_1_1_installer":
+  description: Name of image with 5.1.1 manager installer for bootstrap or cluster installs.
   default: cfy-installer-premium-5.1.1
-5_1_0_installer:
+"5_1_1":
+  description: Name of manager image for 5.1.1. This will be used in all manager tests.
```
YoniItzhak, line 12: I think "master" is the one that will be used in all manager tests. If yes, then "This will be used in all manager tests." should be deleted.

mabuaisha: Yes updated

cosmo_tester/config_schemas/manager_image_names_rhel.yaml
mabuaisha: Updated

YoniItzhak (approved):

PR #1118, opened 2021-01-05
RD-921-Add-IDD-restore-test-cluster

glukhman:


cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -65,6 +68,59 @@ def test_full_cluster_names(full_cluster_names, logger, ssh_key, test_config):
     check_managers(mgr1, mgr2, example)
 
 
+def test_cluster_5_0_5_snapshot_with_idd(full_cluster_ips, logger, ssh_key,
+                                         test_config):
+    broker1, broker2, broker3, db1, db2, db3, mgr1, mgr2 = full_cluster_ips
+
+    snapshot_id = 'snap_5.0.5_with_component_openstack'
+    _upload_snapshot_from_url(mgr2, logger, snapshot_id)
```
geokala, line 76: How big are these snapshots? If you make sure they have no plugins, etc, we can probably get them small enough to include in the snapshots repo and thus avoid an extra external call which will likely be slower and more error prone than uploading from the test runner's storage.

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -65,6 +68,59 @@ def test_full_cluster_names(full_cluster_names, logger, ssh_key, test_config):
...
+    assert "`getting-started.mc-jboss` contains the current deployment" in str(
+        e.value)
+
+    with pytest.raises(CloudifyClientError) as e:
+        mgr1.client.executions.start('infra-openstack', 'uninstall')
+    assert "Can't execute workflow `uninstall` on deployment " \
+           "infra-openstack" in str(e.value)
+    assert "`getting-started.mc-jboss` uses capabilities" in str(e.value)
+    assert "`getting-started.mc-jboss` contains the current deployment" in str(
+        e.value)
```
geokala, line 97: Assuming there is an order in which these deployments should be able to be uninstalled+deleted, this should be checked too.

cosmo_tester/test_suites/cluster/full_cluster_test.py
```diff
@@ -65,6 +68,59 @@ def test_full_cluster_names(full_cluster_names, logger, ssh_key, test_config):
...
+    with pytest.raises(CloudifyClientError) as e:
+        mgr1.client.deployments.delete('infra')
+    assert "Can't delete deployment infra" in str(e.value)
+    assert "`capable` uses capabilities" in str(e.value)
+
+    with pytest.raises(CloudifyClientError) as e:
+        mgr1.client.executions.start('infra', 'uninstall')
+    assert "Can't execute workflow `uninstall` on deployment " \
+           "infra" in str(e.value)
+    assert "`capable` uses capabilities" in str(e.value)
```
geokala, line 121: Aside from the snapshot name and the content of the error messages, these look to be basically doing the same thing, so maybe they could be factored out to make it easier to maintain if we change the test.

cosmo_tester/test_suites/cluster/full_cluster_test.py
geokala: Another alternative would be to add some lightweight (no agents, etc) IDD stuff to the existing snapshot test, rather than adding two extra tests. This would ensure it was working from any version to current.
I think we already have some blueprints that could do this?

geokala (approved):

PR #1137, opened 2021-01-24
RD-1143 Update upgrade system tests to v5.1.2

YoniItzhak:


cosmo_tester/config_schemas/cfy_cluster_manager.yaml
```diff
@@ -8,3 +8,6 @@ manager_install_rpm_path:
 5_1_0_manager_install_rpm_path:
   description: manager-install v5.1.0 RPM path to install.
   default: 'https://repository.cloudifysource.org/cloudify/5.1.0/ga-release/cloudify-manager-install-5.1.0ga.rpm'
+5_1_1_manager_install_rpm_path:
+  description: manager-install v5.1.0 RPM path to install.
```
mabuaisha, line 12: Update description to 5.1.1 instead of 5.1.0

cosmo_tester/framework/util.py
```diff
@@ -658,3 +659,53 @@ def update_dictionary(dict1, dict2):
...
+    """Get the manager-install RPM version
+
+    :param host: A host with cloudify-manager-install RPM installed
+    :return: The cloudif-manager-install RPM version
+    """
+    version = host.run_command(
+        'rpm --queryformat "%{VERSION}" -q cloudify-manager-install',
+        hide_stdout=True)
+
+    return version.stdout
```
mabuaisha, line 711: i think there is a command to that expose the version using `cfy_manager` we can use it instead `cfy_manager version`

cosmo_tester/test_suites/cluster/cfy_cluster_manager_test.py
```diff
@@ -79,24 +81,11 @@ def test_create_three_nodes_cluster(three_vms, three_nodes_config_dict,
...
-        elif i < 6:
-            node_name = 'postgresql-{0}'.format(node_num)
-        else:
-            node_name = 'manager-{0}'.format(node_num)
-
-        nine_nodes_config_dict['existing_vms'][node_name].update({
-            'private_ip': str(node.private_ip_address),
-            'public_ip': str(node.ip_address)
-        })
+    nodes_list = [node for node in nine_vms]
```
mabuaisha, line 84: What is the type of the `nine_vms` ? is not a list ? why do we convert it again ?

cosmo_tester/test_suites/cluster/cfy_cluster_manager_test.py
```diff
@@ -226,6 +215,56 @@ def test_three_nodes_cluster_offline(
                      logger)
 
 
+@pytest.mark.parametrize('base_version', ['5_1_0', '5_1_1'])
+def test_three_nodes_cluster_upgrade(base_version, three_vms,
+                                     three_nodes_config_dict, test_config,
+                                     ssh_key, logger):
+    """Tests the command cfy_cluster_manager upgrade on a 3 nodes cluster."""
+    node1, node2, node3 = three_vms
+    nodes_list = [node1, node2, node3]
```
mabuaisha, line 224: Is there a reason to not use the same list ?

mabuaisha (commented):
left some comments

cosmo_tester/framework/util.py
YoniItzhak: You are right, there is a command for that (I was the one who even created it), but it just logs the version and doesn't return it in a convenient form for parsing. This is why I used this approach.

cosmo_tester/test_suites/cluster/cfy_cluster_manager_test.py
YoniItzhak: `nine_vms` is a generator https://github.com/cloudify-cosmo/cloudify-system-tests/blob/8d5428cb31304fda7837725340449dfcb6ca2964/cosmo_tester/test_suites/cluster/conftest.py#L133, this is why we convert it into a list.

cosmo_tester/test_suites/cluster/cfy_cluster_manager_test.py
YoniItzhak: As `nine_vms`, `three_vms` is also a generator and not a list.

mabuaisha (approved):

PR #1172, opened 2021-04-09
514 fixes

geokala:


geokala:
As 5.1.4 has been released now and this was used to test it, I'll just merge.

PR #1189, opened 2021-05-03
Bounding cryptography

YoniItzhak:
When pip installing cloudify-system-test cryptography v3.4.7 throws the following error 

```
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-sivmumzz/cryptography/setup.py", line 14, in <module>
        from setuptools_rust import RustExtension
    ModuleNotFoundError: No module named 'setuptools_rust'
```

Bounding it to v3.3.2 fixes the issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ chardet==3.0.4
     # via requests
 https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip#egg=cloudify-common[dispatcher]
     # via -r requirements.in
-cryptography==3.4.7
+cryptography==3.3.2
     # via
     #   paramiko
     #   requests-ntlm


requirements.txt
```diff
@@ -25,7 +25,7 @@ chardet==3.0.4
     # via requests
 https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip#egg=cloudify-common[dispatcher]
     # via -r requirements.in
-cryptography==3.4.7
+cryptography==3.3.2
```
geokala, line 28: This should be bounded in the setup.py and then the requirements.txt regenerated. Otherwise, any time we regenerate this file we'll lose this bounding.

geokala:
As a note, this bounding shouldn't be required anyway- you can avoid it IIRC if you "pip install --upgrade pip setuptools"

YoniItzhak:
Following @geokala 's comments, this PR is closed.

PR #1220, opened 2021-05-25
Remove 5.1.0 from upgrade base versions

YoniItzhak:
According to Ofer, we don't support upgrading from v5.1.0 using `cfy_manager upgrade`, so we should also remove it from the tests.

### DIFF
--- a/cosmo_tester/test_suites/cluster/cfy_cluster_manager_shared.py
+++ b/cosmo_tester/test_suites/cluster/cfy_cluster_manager_shared.py
@@ -5,15 +5,9 @@ import pkg_resources
 from retrying import retry
 import yaml
 
-from cosmo_tester.framework.constants import SUPPORTED_RELEASES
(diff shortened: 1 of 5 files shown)


cosmo_tester/test_suites/cluster/cfy_cluster_manager_shared.py
```diff
@@ -12,7 +12,7 @@
 BASE_VERSIONS = [
     version + '-ga'
     for version in SUPPORTED_RELEASES
-    if version not in ('master', '5.0.5')
+    if version not in ('master', '5.0.5', '5.1.0')
```
geokala, line 15: The same change is needed in cosmo_tester/test_suites/general/cfy_manager_test.py I think. Maybe factor both out into constants.py as a SUPPORTED_FOR_RPM_UPGRADE?

geokala (approved):

PR #1222, opened 2021-05-25
Fix removing service in tests

YoniItzhak:


geokala:
Note: This is to work around a c-m-i issue which won't be fixed for 6.0.0.

PR #1232, opened 2021-05-27
RD-858 test_inplace_restore: wait_for_manager after snap-restore

tehasdf:
After restoring the snapshot, we wait for the manager to come back
up, because the snapshot-restore includes a reboot.

Only THEN we wait for agents. Why would we wait for agents before
the manager was up?

The order of operations is:
1. snap-restore
2. reboot
3. wait for manager to come back up
4. wait for agents to reconnect

Before this change, 3 and 4 were swapped


tehasdf:
commit 2 also reduces the wait time

geokala (approved):

PR #1242, opened 2021-06-08
Faster summary

geokala:


geokala:
Side note: This should save us most of a dollar for each full test run we do.

tehasdf (approved):

PR #1254, opened 2021-06-17
RD-2464 Update tests who use full_clester_ips fixture

glukhman:


cosmo_tester/test_suites/cluster/certificate_replacement_cluster_tests.py
```diff
@@ -7,7 +7,7 @@
 
 def test_replace_certificates_on_cluster(full_cluster_ips, logger, ssh_key,
                                          test_config, module_tmpdir):
-    broker1, broker2, broker3, db1, db2, db3, mgr1, mgr2 = full_cluster_ips
+    broker1, broker2, broker3, db1, db2, db3, mgr1, mgr2, _ = full_cluster_ips
```
geokala, line 10: Probably best to put it as mgr3- we've got this in other files, I must've missed this one.

cosmo_tester/framework/test_hosts.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, image_type, test_config, bootstrappable=False):
         self.api_ca_path = None
         self.enable_ssh_wait = True
         self.should_finalize = True
+        self.should_teardown = True
```
geokala, line 51: This won't work, because this means that we'll tear down just the db service and then mark the node as not needing to tear down. The 'should_teardown' flag is essentially the installed_configs list, so something needs to be done with that.

cosmo_tester/test_suites/cluster/db_management_test.py
```diff
@@ -5,7 +5,7 @@
 
 
 def test_remove_db_node(full_cluster_ips, logger, ssh_key, test_config):
-    broker1, broker2, broker3, db1, db2, db3, mgr1, mgr2 = full_cluster_ips
+    broker1, broker2, broker3, db1, db2, db3, mgr1, mgr2, _ = full_cluster_ips
```
geokala, line 8: This one should also be mgr3

geokala (approved):

PR #1269, opened 2021-07-05
RD-2682 Explicitly use new generated CA cert in client before replacing

glukhman:


cosmo_tester/test_suites/general/aio_replace_certs_test.py
```diff
@@ -17,6 +17,10 @@ def test_aio_replace_certs(image_based_manager, ssh_key, logger, test_config):
                                       replace_certs_config_path,
                                       ssh_key.private_key_path)
 
+    local_new_ca_path = join(str(module_tmpdir), 'new_ca.crt')
```
geokala, line 20: It might be better to add a force flag to https://github.com/cloudify-cosmo/cloudify-system-tests/blob/master/cosmo_tester/framework/test_hosts.py#L689 then just call that, because then you don't need to update the cert setting (and you can do the same change in the cluster certs replace test).

geokala (approved):

PR #1286, opened 2021-07-09
RD-2425 IPv6 cluster test

mateumann:


cosmo_tester/framework/test_hosts.py
```diff
@@ -1105,7 +1115,9 @@ def _upload_blueprints_to_infrastructure_manager(self):
         self._logger.info(
             'Uploading test blueprints to infrastructure manager.'
         )
-        suffix = '-multi-net' if self.multi_net else ""
+        suffix = '-multi-net' if self.multi_net \
```
geokala, line 1118: How about just appending each of the possible suffixes so that making an ipv6 multi-net test later (if we decided to) would just be a matter of adding the blueprints and removing the check you put in above.

geokala: (I think it'll also make this logic look a bit cleaner)

cosmo_tester/framework/test_hosts.py
```diff
@@ -1369,3 +1390,21 @@ def _update_instance(self, server_index, node_instance):
             server_id,
             server_index,
         )
+
+    def _disable_ipv4(self, node_instances):
```
geokala, line 1394: This should be on the host itself in the finalise, then you don't need to separately wait_for_ssh, because you can put it on after that, and just make it auto-execute this if the test VM is marked as ipv6.

cosmo_tester/resources/infrastructure_blueprints/aws/vm-ipv6-net.yaml
```diff
@@ -0,0 +1,148 @@
...
+        target: nic
+
+  subnet:
+    type: cloudify.nodes.aws.ec2.Subnet
+    properties:
+      use_external_resource: true
+      resource_id: { get_input: subnet_id }
+      client_config: *client_config
+      resource_config:
+        CidrBlock: 192.168.42.0/24
```
geokala, line 99: Why do you have an IPv4 block in your IPv6 blueprint? I'm assuming it's because of the floating IPs, but maybe add a comment to that effect if that's right?

cosmo_tester/framework/test_hosts.py
mateumann: Do you mean this part of code https://github.com/cloudify-cosmo/cloudify-system-tests/blob/f8fe18688e683375208adc48d9c452eb4ef4f39b/cosmo_tester/framework/test_hosts.py#L244-L245?  I am not sure we know IP addresses of other two VMs at this point.

cosmo_tester/resources/infrastructure_blueprints/aws/vm-ipv6-net.yaml
mateumann: Yup, it seems some form of IPv4 is necessary for the floating/elastic IP to work.  Comment added

cosmo_tester/framework/test_hosts.py
geokala: Ah, I see the problem then, yes. Fair enough. Can you add a comment noting that that is done in the test_hosts because the IPs of all hosts need to be known before applying the settings?

cosmo_tester/resources/infrastructure_blueprints/aws/vm-ipv6-net.yaml
geokala: That's a pity but let's not worry about it more right now.

cosmo_tester/framework/test_hosts.py
```diff
@@ -1115,9 +1115,13 @@ def _upload_blueprints_to_infrastructure_manager(self):
         self._logger.info(
             'Uploading test blueprints to infrastructure manager.'
         )
-        suffix = '-multi-net' if self.multi_net \
-            else '-ipv6-net' if self.ipv6_net \
-            else ''
+        suffix = ''
+        if self.ipv6_net:
```
geokala, line 1119: Let's make them -ipv6-net and -multi-net? That way if we do an ipv6 multi-net later it'll just work (if we name blueprints correctly).

cosmo_tester/test_suites/cluster/compact_cluster_test.py
```diff
@@ -66,6 +66,16 @@ def test_three_nodes_cluster_teardown(three_nodes_cluster, ssh_key,
     _assert_cluster_status(node1.client)
 
 
+@pytest.mark.three_vms_ipv6
+def test_three_nodes_cluster_ipv6_status(three_nodes_ipv6_cluster, logger):
+    node1, node2, node3 = three_nodes_ipv6_cluster
```
geokala, line 71: We should probably be deploying an agent as well (which I realise will make this not really just be a _status test, so maybe drop that suffix), using the ipv6 address of one of the managers.

geokala (dismissed):

geokala (approved):

PR #1310, opened 2021-07-18
Rd 2754 fix cluster install tests

glukhman:


cosmo_tester/test_suites/general/aio_replace_certs_test.py
```diff
@@ -17,6 +17,10 @@ def test_aio_replace_certs(image_based_manager, ssh_key, logger, test_config):
                                       replace_certs_config_path,
                                       ssh_key.private_key_path)
 
+    local_new_ca_path = join(str(module_tmpdir), 'new_ca.crt')
+    image_based_manager.get_remote_file('~/.cloudify-test-ca/ca.crt', local_new_ca_path)
+    image_based_manager.client._client.cert = local_new_ca_path
```
geokala: This would be better using the image_based_manager.download_rest_ca(force=True) and image_based_manager.client = image_based_manager.get_rest_client() after the certs replace operation runs.

geokala: Whoops, no need for me to comment that, it happened in the next commit.

cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -276,6 +278,8 @@ def _ensure_installer_not_installed(vm):
             ]),
         )
     )
+    vm.run_command('if rpm -qi cloudify-manager-install; then '
```
geokala, line 281: This is unnecessary- cloudify-manager-install is removed in the previous command.

cosmo_tester/test_suites/cluster/conftest.py
```diff
@@ -240,7 +241,8 @@ def nine_vms(nine_session_vms, test_config, logger):
                      broker_count=3, db_count=3,
                      manager_count=3, bootstrap=False)
     for vm in nine_session_vms:
-        vm.teardown()
+        _remove_cluster(vm, logger)
+        vm.teardown(kill_certs=False)
```
geokala, line 245: Certs should be killed, there shouldn't be relation required between the certs of different tests.

PR #1316, opened 2021-07-21
RD-2788 Restart on inplace restore to avoid race

geokala:


geokala:
Marking these as draft while I double check because the findings don't feel entirely correct.

PR #1317, opened 2021-07-22
add region to clean-up script

yuvlar:


geokala (approved):
I note that we're defaulting this manually here, which may cause problems if we later change the default in the config_schemas, but we can make this script better if that time comes.

PR #1327, opened 2021-07-26
Rd 2754 fix cluster upgrade test

glukhman:


glukhman:
should be in cluster manager instead, if needed at all, following https://github.com/cloudify-cosmo/cloudify-manager-install/pull/1204

PR #1374, opened 2022-01-03
Clean up logs dir on teardown

geokala:
Leaving the logs dir in place between runs causes permissions issues
when using older versions of cfy_manager after newer ones.


mateumann (approved):

geokala:
OK, this isn't quite fixing it. I'll investigate more.

geokala:
Not needed since rsync merged.

PR #1395, opened 2022-01-16
No basic_install_config for non-managers

glukhman:


geokala:
We appear to have already merged this change in another PR.

PR #1419, opened 2022-02-28
Randomise passwords on test VMs

geokala:
Otherwise we put publically accessible VMs up with default passwords in most cases.

### DIFF
--- a/cosmo_tester/framework/util.py
+++ b/cosmo_tester/framework/util.py
@@ -747,6 +747,6 @@ def rsync_restore(nodes):
 
 def generate_password():
     charset = ascii_uppercase + ascii_lowercase + digits
-    return ''.join([
+    return ''.join(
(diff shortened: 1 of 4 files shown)


tehasdf (dismissed):

cosmo_tester/framework/util.py
```diff
@@ -741,3 +743,10 @@ def rsync_restore(nodes):
             time.sleep(3)
         node.log_action('Rsync restore complete')
         node.reboot_required = True
+
+
+def generate_password():
+    charset = ascii_uppercase + ascii_lowercase + digits
+    return ''.join([
+        random.choice(charset) for _ in range(30)
+    ])
```
tehasdf, line 752:
```suggestion
    return ''.join(
        random.choice(charset) for _ in range(30)
    )
```

tehasdf (approved):

PR #1443, opened 2022-06-19
RD-2705 Add certs test for 3-node cluster

glukhman:


cosmo_tester/resources/scripts/create_replace_certs_config_script.py
```diff
@@ -27,19 +29,23 @@ def generate_replace_certs_config(replace_certs_config_path,
...
             key_path = certs_dir + host_ip + '.key'
             for cert_name in instance_dict:
-                if 'ca' in cert_name:
+                if 'ca' in cert_name and 'key' not in cert_name:
                     instance_dict[cert_name] = ca_path
                 elif 'cert' in cert_name:
                     instance_dict[cert_name] = cert_path
                 elif 'key' in cert_name:
                     instance_dict[cert_name] = key_path
+                elif replace_ca_key and 'ca_key' in cert_name:
```
geokala, line 47: You'll never hit this branch, it'll always land on the previous one.

I suggest instead making the first branch handle all ca stuff- something like nesting below if 'ca' in cert_name: if 'ca_key' in cert_name: <something> else: <other thing>

geokala (approved):

PR #1451, opened 2022-07-21
RD-5408 Allow Cloudify versions with no RH8 support spin up with RH7 …

glukhman:
…instead


cosmo_tester/framework/test_hosts.py
```diff
@@ -904,9 +905,18 @@ def _is_manager_image_type(self):
...
+            return True
+        if parse_version(self.image_type) >= parse_version('6.4.0'):
+            return True
+        return False
+
     def _set_image_details(self):
         if self.is_manager:
             distro = self._test_config['test_manager']['distro']
+            if distro == 'rhel-8' and not self._is_rhel8_supported():
+                distro == 'rhel-7'
```
geokala, line 920: Don't you also need to do something to set the image name further down if it is/n't RHEL8?

PR #1457, opened 2022-07-26
Rd 5438 rh8 systest fixes

glukhman:


cosmo_tester/framework/test_hosts.py
```diff
@@ -882,6 +882,8 @@ def enable_nics(self):
 
         self._logger.info('Adding extra NICs...')
 
+        self.run_command('yum install -y ifupdown', use_sudo=True)
```
geokala, line 885: For this and the other yum install, they shouldn't be in the framework, they should be installed on the image that needs to use them.

geokala (approved):

PR #1484, opened 2022-09-06
RD-5243 Test for broken agent/manager connection

mateumann:


geokala (approved):
This is in 6.4.1. It does require a fix, but I'm going to green and merge this so that the same fix can be applied here.

PR #1513, opened 2022-12-05
Rd 6504 test 700 inplace upgrade

glukhman:


cosmo_tester/framework/constants.py
```diff
@@ -9,6 +9,8 @@
     '6.0.0',
     '6.1.0',
     '6.2.0',
+    '6.3.0',
+    '6.4.0',
```
geokala, line 13: Should be 6.4.1 as well- first and last of each minor version.

Can you ask Alex if we can remove 5.0.5 and 5.1.x too? (maybe even 5.2.0 if not 5.2.7)

PR #1521, opened 2023-01-11
Rd 6678 test upgrade ext db

glukhman:


cosmo_tester/test_suites/cluster/external_component_cluster_test.py
```diff
@@ -32,3 +38,66 @@ def test_cluster_single_db(cluster_with_single_db, logger, ssh_key,
...
     check_managers(mgr1, mgr2, example)
+
+
+@pytest.mark.three_vms
+@pytest.mark.upgrade
+@pytest.mark.parametrize('base_version', ['6.3.2-ga', '6.4.1-ga'])
+def test_upgrade_external_db(
+        base_version, three_vms, logger, ssh_key, test_config):
+    ext_db_fqdn = "test-db-pg10.cqezfgjnjobz.eu-west-1.rds.amazonaws.com"
+    ext_db_password = "a1234567"
```
mateumann, line 49: This doesn't look great.  These lines suggest, that there is always-on DB instance, which can be accessed using these globally readable credentials.  Is that so?  Are we OK with allowing anyone the access to the DB which we pay for?

mateumann (approved):

PR #1525, opened 2023-02-06
RD-6658 agent systests: the system name is "linux"

tehasdf:
"manylinux" isn't a system, but "linux" is.

### DIFF
--- a/cosmo_tester/test_suites/agent/__init__.py
+++ b/cosmo_tester/test_suites/agent/__init__.py
@@ -65,6 +65,6 @@ def validate_agent(manager, example, test_config,
 
     agent_system = agent.get('system')
     assert agent_system == expected_system or \
-           agent_system.startswith('manylinux')
+           agent_system.startswith('linux')
     for key in expected_agent:
         assert agent.get(key) == expected_agent[key]


adrian-polanczyk-codilime (approved):

cosmo_tester/test_suites/agent/__init__.py
```diff
@@ -65,6 +65,6 @@ def validate_agent(manager, example, test_config,
 
     agent_system = agent.get('system')
     assert agent_system == expected_system or \
-           agent_system.startswith('manylinux')
+           agent_system.startswith('linux')
```
geokala, line 68: Hmmm... I think this might get used with older versions of managers as well. Did we do manylinux here prior to 7?

tehasdf: we have not: https://github.com/cloudify-cosmo/cloudify-manager/blob/6.4.2-build/packaging/cloudify-agents.spec#L33-L40

([master](https://github.com/cloudify-cosmo/cloudify-manager/blob/9d17950569e091dbeb43fda6fda63c014b9ee661/packaging/cloudify-agents.spec#L33-L35))

geokala (approved):

PR #1526, opened 2023-02-09
RD-6996 Make AIO upgrade test happy

geokala:


geokala:
Ignore the branch name, too much talk about agent issues. Incidentally, this test does still fail due to an agent upgrade issue, which is under discussion separately.

tehasdf (approved):

